PR #19, opened 2020-05-29
Dataset Schema 0.1.0

adason:
# Peer Review Information:
This change propose update to dataset schema that decouples annotations from the captures.json files itself. Instead annotations are stored in a separate file to reduce memory requirements loading `captures.json`.

## Dev Testing:
**Tests Added**: 
<br>
<br>
**Core Scenario Tested**: 
<br>
**At Risk Areas**: 
<br>
**Notes + Expectations**:


adason:
This change was required to reduce memory footprint in `captures`. Using bag can filter annotations from captures, but the annotations has to be loaded into memory later in order to train models. We could delay this loading but it's hard to break annotations and index properly. 

The values stored in annnotations json are only required when we are loading images/annotations into mini-batches for training. Filenames are served as simple "index" to allow users pointing a given capture_id to the right image/annotation files. For similar reason, we never need to load all images binaries to memory in order to train ML models.

I agree this will create 2X overhead to store extra files for annotations, as one will need one such file per image. The alternative was to store annotations in more efficient local filesystem but that does require a separate pipeline simply extract annotations and stored in more efficient format. Since we have full control of SDK, I think it's better to handle the file splitting upfront instead of asking datasetinsights to add extra processing steps to split annotation values.

sleal-unity:
I think this PR has been open for months now with no progress, so I'm gonna go ahead and close it.

If we need to address this again, we can always reopen it.

adason:
I want to re-open this PR to make progress on decoupling annotation json files.

adason:
Moved discussion to https://github.com/Unity-Technologies/com.unity.perception/pull/337

PR #25, opened 2020-06-18
Publish fix for Yamato CI

Wesley201:
# Peer Review Information:
Pulled the Publish step into its own file and dependent on the package tests, errors appear to be fixed 

2019.3.15f1

## Dev Testing:
**Tests Added**: 
<br>
<br>
**Core Scenario Tested**: 
CI green
<br>
**At Risk Areas**: 
Still working out the trigger process for the publish step
<br>
**Notes + Expectations**:


.yamato/publish.yml
```diff
@@ -0,0 +1,29 @@
+{% metadata_file .yamato/environments.yml %}
+
+---
+{% for variant in package_variants %}
```
JonathanHUnity, line 4: We can do away with the variants now - I can't think of any use for variants in the Perception package.

PR #31, opened 2020-07-02
Checks and doc updates for URP RenderFeature requirement

JonathanHUnity:
# Peer Review Information:
Perception on URP requires that the user add a GroundTruthRenderFeature to the UniversalRenderPipelineAsset. This PR adds instructions for doing this to the manual and adds a console error at runtime when the GroundTruthRenderFeature isn't present.

## Dev Testing:
**Tests Added**:
Added tests around new error messages.
<br>
<br>
**Core Scenario Tested**: 
<br>
**At Risk Areas**: 
<br>
**Notes + Expectations**:


com.unity.perception/Runtime/GroundTruth/PerceptionCamera.cs
```diff
@@ -74,6 +74,10 @@ public void AddScriptableRenderPass(ScriptableRenderPass pass)
         static ProfilerMarker s_FlipY = new ProfilerMarker("Flip Y (PerceptionCamera)");
         static ProfilerMarker s_EncodeAndSave = new ProfilerMarker("Encode and save (PerceptionCamera)");
 
+#if URP_PRESENT
+        bool m_GroundTruthRendererFeatureRun;
```
idanbeck, line 78: Not sure what the convention is, but I generally use f for flags as in m_fGroundTruthRendererFeatureRun, and I've seen is/has etc throughout code before.  

Also, does it make sense to put this in an IFDEF vs. using a nullable variable for example?  Not a strong opinion here - since it's just a flag.

com.unity.perception/Runtime/GroundTruth/PerceptionCamera.cs
```diff
@@ -104,9 +108,25 @@ void Awake()
             SetupInstanceSegmentation();
 
             RenderPipelineManager.beginCameraRendering += OnBeginCameraRendering;
+            RenderPipelineManager.endCameraRendering += CheckForRendererFeature;
             DatasetCapture.SimulationEnding += OnSimulationEnding;
         }
 
+
+        void CheckForRendererFeature(ScriptableRenderContext arg1, Camera arg2)
```
idanbeck, line 116: arg1, arg2 are pretty generic names when they're used in a specific context in the actual function.  I think they should be renamed to even something like scriptableRC, and camera respectively.

com.unity.perception/Runtime/GroundTruth/PerceptionCamera.cs
```diff
@@ -286,5 +306,12 @@ public bool RemoveLabeler(CameraLabeler cameraLabeler)
             }
             return false;
         }
+
+#if URP_PRESENT
```
idanbeck, line 310: Sort of mirroring the above point - having a function of a class per an ifdef will result in inflexible arch down the road.  I think it'd be better to at least put the IFDEF inside and then call up to the parent explicitly if this is an override. 

Does C# have an explicit override keyword like C++?  If so I think that could also make this a bit more clear

com.unity.perception/Runtime/GroundTruth/PerceptionCamera.cs
JonathanHUnity: The m_ is per our standard conventions for private members. Besides m_/k_ we aren't actually doing hungarian notation :) 
I've removed the IFDEFs everywhere except the final check. I agree it's cleaner this way

com.unity.perception/Runtime/GroundTruth/PerceptionCamera.cs
JonathanHUnity: Fixed!

com.unity.perception/Runtime/GroundTruth/PerceptionCamera.cs
JonathanHUnity: Fixed. This isn't overriding anything in the base class though, so I think it doesn't apply.

PR #40, opened 2020-07-20
Aisv647 visualizations - Preview Cut

StevenBorkman:
# Peer Review Information:
Preview cut of the perception visualizers

## Dev Testing:
**Tests Added**: 
<br>
<br>
**Core Scenario Tested**: 
<br>
**At Risk Areas**: 
<br>
**Notes + Expectations**:


com.unity.perception/Runtime/GroundTruth/Labelers/BoundingBoxLabeler.cs
```diff
@@ -108,8 +116,73 @@ void OnRenderedObjectInfosCalculated(int frameCount, NativeArray<RenderedObjectI
                     };
                 }
 
+                if (!CaptureOptions.useAsyncReadbackIfSupported && frameCount != Time.frameCount)
```
JonathanHUnity, line 119: Looks like the current boxes are not in sync with the frame. I guess it's because we aren't reading back synchronously while the UI is up?

com.unity.perception/Runtime/GroundTruth/Labelers/BoundingBoxLabeler.cs
```diff
@@ -108,8 +116,73 @@ void OnRenderedObjectInfosCalculated(int frameCount, NativeArray<RenderedObjectI
...
+            visualizationHolder.transform.SetParent(panel.transform.parent, false);
+            
+            visualizationPanelCache = new Dictionary<string, RectTransform>();
+        }
+
+        void Visualize()
+        {
+            foreach (var box in m_BoundingBoxValues)
+            {
+                var rectTrans = GetVisualizationPanel(box.label_name);
```
JonathanHUnity, line 152: This appears to assume that each label will only ever have one bounding box, which is not the case.

com.unity.perception/Runtime/GroundTruth/Labelers/BoundingBoxLabeler.cs
```diff
@@ -108,8 +116,73 @@ void OnRenderedObjectInfosCalculated(int frameCount, NativeArray<RenderedObjectI
                     };
                 }
 
+                if (!CaptureOptions.useAsyncReadbackIfSupported && frameCount != Time.frameCount) 
+                    Debug.LogWarning("Not on current frame: " + frameCount + "(" + Time.frameCount + ")");
+
+                if (IsVisualizationEnabled()) 
+                {
+                    Visualize();
+                    ReportFrameCount(frameCount);
```
JonathanHUnity, line 125: Probably the 2d bounding box labeler is not the right place to report the frame count.

com.unity.perception/Runtime/GroundTruth/Labelers/CameraLabeler.cs
```diff
@@ -34,6 +51,15 @@ public abstract class CameraLabeler
...
         protected virtual void Setup() { }
         /// <summary>
+        /// Called immediately after <see cref="setup"/>. Implement this to initialize labeler's visualization
+        /// capability if one exists <see cref="supportVisualization"/>.
+        /// </summary>
+        protected virtual void SetupVisualizationPanel(GameObject panel) { }
+        /// <summary>
+        /// Called when the labeler's visualization capability is turned on or off.
+        /// </summary>
+        protected virtual void OnVisualizerEnabled(bool enabled) {}
```
JonathanHUnity, line 61: The name implies (to me) that the visualizer is being enabled, but actually it is either being enabled or disabled. Consider changing to `OnVisualizer<something>Changed`

com.unity.perception/Runtime/GroundTruth/Labelers/CameraLabeler.cs
```diff
@@ -48,10 +74,47 @@ public abstract class CameraLabeler
...
         internal void InternalOnUpdate() => OnUpdate();
         internal void InternalOnBeginRendering() => OnBeginRendering();
         internal void InternalCleanup() => Cleanup();
 
+        /// <summary>
+        /// Turns on/off the labeler's realtime visualization capability. If a labeler
+        /// does not support realtime visualization (<see cref="supportsVisualization"/>)
+        /// this will not function.
+        /// </summary>
+        protected void EnableVisualization(bool enabled)
```
JonathanHUnity, line 88: Consider a property here.

com.unity.perception/Runtime/GroundTruth/Labelers/BoundingBoxLabeler.cs
```diff
@@ -108,8 +116,73 @@ void OnRenderedObjectInfosCalculated(int frameCount, NativeArray<RenderedObjectI
...
+        {
+            var toggle  = GameObject.Instantiate(Resources.Load<GameObject>("GenericToggle"));
+            toggle.transform.SetParent(panel.transform);
+            toggle.GetComponentInChildren<Text>().text = "Bounding Boxes";
+            toggle.GetComponent<Toggle>().onValueChanged.AddListener(enabled => {
+                EnableVisualization(enabled);
+            });
+
+            visualizationHolder = GameObject.Instantiate(Resources.Load<GameObject>("BoundsHolder"));
+            visualizationHolder.transform.SetParent(panel.transform.parent, false);
```
JonathanHUnity, line 143: `panel.transform.parent` is used in a couple places as the container for visualizations. Should this be exposed explicitly?

com.unity.perception/Runtime/GroundTruth/Labelers/BoundingBoxLabeler.cs
```diff
@@ -108,8 +116,73 @@ void OnRenderedObjectInfosCalculated(int frameCount, NativeArray<RenderedObjectI
...
+                }
+                
                 asyncAnnotation.ReportValues(m_BoundingBoxValues);
             }
         }
+
+        /// <inheritdoc/>
+        protected override void SetupVisualizationPanel(GameObject panel)
+        {
+            var toggle  = GameObject.Instantiate(Resources.Load<GameObject>("GenericToggle"));
```
JonathanHUnity, line 135: Thinking about the experience from a user perspective - will a user be expected to use Resources.Load() to instantiate UI prefabs? I'm a bit wary of asking users to depend on the presense and internals of specific prefabs in the package. Maybe we could wrap this with a little factory class and then give users the freedom to add their own UI by hand?

com.unity.perception/Runtime/GroundTruth/Labelers/Resources/BoundingBoxPrefab.prefab
```diff
@@ -0,0 +1,171 @@
+%YAML 1.1
```
JonathanHUnity, line 1: Regarding the use of the Resources folder in packages, there was an edict (so to speak) that packages should (almost) never use the Resources folder to prevent adding unnecessary data to players. I don't think this is a huge deal for Perception, since a few bytes in these players is not that big of a deal, but it's something to read through and consider.

https://docs.google.com/document/d/1sb3SMQHBQMWvPifgabLK40pvptR01bfJeMX9GxRnw28/edit

com.unity.perception/Runtime/GroundTruth/Labelers/Resources/KeyValuePanel.cs
```diff
@@ -0,0 +1,39 @@
...
+    ///
+    /// </summary>
+    public class KeyValuePanel : MonoBehaviour
+    {
+        Text key = null;
+        Text value = null;
+
+        void Awake()
+        {
+            key = this.transform.Find("Key").GetComponent<Text>();
```
JonathanHUnity, line 18: Instead of Find, these references can be serialized in the panel. Or, you could create the objects at runtime. In its current state, this type has some very weakly expressed contracts around its child objects.

com.unity.perception/Runtime/GroundTruth/Labelers/SemanticSegmentationLabeler.cs
```diff
@@ -167,6 +177,9 @@ protected override void Setup()
 
             m_SemanticSegmentationTextureReader = new RenderTextureReader<Color32>(targetTexture, myCamera,
                 (frameCount, data, tex) => OnSemanticSegmentationImageRead(frameCount, data));
+        
+            supportsVisualization = true;
+            EnableVisualization(supportsVisualization);
```
JonathanHUnity, line 182: Should this be called automatically?

com.unity.perception/Runtime/GroundTruth/Labelers/CameraLabeler.cs
```diff
@@ -60,12 +123,61 @@ internal void Init(PerceptionCamera newPerceptionCamera)
...
             }
         }
+
+        private GameObject GetCanvas()
+        {
+            var canvas = GameObject.Find("Canvas");
+            
+            if (canvas == null)
+            {
+               canvas = GameObject.Instantiate(Resources.Load<GameObject>("Canvas"));
```
JonathanHUnity, line 145: As we found with HDRP, interacting with the canvas requires an EventManager. Can we automatically create an EventSystem if it does not already exist?

com.unity.perception/Runtime/GroundTruth/Labelers/Resources/KeyValuePanel.cs
```diff
@@ -0,0 +1,39 @@
...
+using System.Collections.Generic;
+using UnityEngine;
+using UnityEngine.UI;
+
+namespace UnityEngine.Perception.GroundTruth
+{
+    /// <summary>
+    ///
+    /// </summary>
+    public class KeyValuePanel : MonoBehaviour
```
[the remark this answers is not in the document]
StevenBorkman: For sure, should we create a visualization folder under labelers to hold visualization scripts?

com.unity.perception/Runtime/GroundTruth/RenderTextureReader.cs
```diff
@@ -34,8 +34,7 @@ public RenderTextureReader(RenderTexture source, Camera cameraRenderingToSource,
             this.m_CameraRenderingToSource = cameraRenderingToSource;
             m_NextFrameToCapture = Time.frameCount;
 
-            if (!GraphicsUtilities.SupportsAsyncReadback())
-                m_CpuTexture = new Texture2D(m_Source.width, m_Source.height, m_Source.graphicsFormat, TextureCreationFlags.None);
+            m_CpuTexture = new Texture2D(m_Source.width, m_Source.height, m_Source.graphicsFormat, TextureCreationFlags.None);
```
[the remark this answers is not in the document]
StevenBorkman: It's hard to know when someone may choose to go into visualization mode, and right now, we are creating the cpu texture as a part of initialization. So either we need to create the cpu texture and take the chance that it may never be used, or check if it has been allocated before we use it each time.

com.unity.perception/Runtime/GroundTruth/Labelers/CameraLabeler.cs
```diff
@@ -60,12 +123,61 @@ internal void Init(PerceptionCamera newPerceptionCamera)
...
             catch (Exception)
             {
                 this.enabled = false;
                 throw;
             }
         }
+
+        private GameObject GetCanvas()
+        {
+            var canvas = GameObject.Find("Canvas");
```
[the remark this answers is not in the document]
StevenBorkman: I was doing this in case some user was using a scene that already had a canvas object as a part of it. Also, the use case that multiple visualizers are being used, we don't want to have multiple canvases being created (although that could be solved using a static canvas variable). I agree that we should avoid using GameObject.Find() because of performance reasons. But in this case it should only be called once during initialization.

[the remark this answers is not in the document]
StevenBorkman: Added a cached member to the CameraLabeler class to make sure that the `GameObject.Find()` is only called at most once per visualizer at init time.

com.unity.perception/Runtime/GroundTruth/Labelers/CameraLabeler.cs
```diff
@@ -17,6 +18,22 @@ public abstract class CameraLabeler
 
         internal bool isInitialized { get; private set; }
 
+        /// <summary>
+        /// Labelers should set this in their setup to define if they support realtime
+        /// visualization of their data. 
+        /// </summary>
+        protected bool supportsVisualization = false;
```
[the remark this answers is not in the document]
StevenBorkman: Good idea. Implemented.

com.unity.perception/Runtime/GroundTruth/Labelers/Resources/KeyValuePanel.cs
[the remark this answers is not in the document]
JonathanHUnity: Makes sense to me. Also a new namespace would be good.

com.unity.perception/Runtime/GroundTruth/RenderTextureReader.cs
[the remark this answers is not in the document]
JonathanHUnity: At the moment this type still only uses `GraphicsUtilities.SupportsAsyncReadback()` to determine whether it should use m_CpuTexture. Were you intending some other changes here to go along with the visualization toggle?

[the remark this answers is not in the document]
StevenBorkman: Changed it to check if `m_CpuTexture == null` and allocating it before accessing it.

com.unity.perception/Runtime/GroundTruth/Labelers/CameraLabeler.cs
```diff
@@ -48,10 +74,47 @@ public abstract class CameraLabeler
...
+                    CaptureOptions.useAsyncReadbackIfSupported = true;
+
+                OnVisualizerEnabled(enabled);
+            }
+        }
+
+        /// <summary>
+        /// Is the visualization capability of the labeler currently active.
+        /// </summary>
+        protected bool IsVisualizationEnabled()
```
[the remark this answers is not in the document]
StevenBorkman: Updated

com.unity.perception/Runtime/GroundTruth/Labelers/CameraLabeler.cs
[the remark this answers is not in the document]
StevenBorkman: That's how we determine if we are going to create the UI elements for the labeler. We could find a different way to do this probably, but I'm not sure if it will gain us anything. Maybe?

com.unity.perception/Runtime/GroundTruth/Labelers/BoundingBoxLabeler.cs
StevenBorkman: So I put a warning logging statement in if we are rendering results from a different frame than the one that we are currently on, and then doesn't log, so I don't know why. I think (although I'm not 100% sure anymore) that I was more in sync with Priyesh's fixes, and being a frame off is definitely something that he had mentioned. So I was putting all of my eggs in the basket of his fix fixing it. If not then some fun debugging.

com.unity.perception/Runtime/GroundTruth/Labelers/BoundingBoxLabeler.cs
StevenBorkman: changed to use `GetVisualizationPanel(box.label_name + "_" + box.instance_id);`

com.unity.perception/Runtime/GroundTruth/Labelers/BoundingBoxLabeler.cs
StevenBorkman: Yeah, that was something that I put in for testing and forgot to remove. It's gone now.

com.unity.perception/Runtime/GroundTruth/Labelers/BoundingBoxLabeler.cs
JonathanHUnity: I don't think so. Can you also repro this?

com.unity.perception/Runtime/GroundTruth/Labelers/CameraLabeler.cs
StevenBorkman: Changed to `OnVisualizerActiveStateChanged()` <- It's kind of lame, but 🤷‍♂️

com.unity.perception/Runtime/GroundTruth/Labelers/CameraLabeler.cs
StevenBorkman: Done

com.unity.perception/Runtime/GroundTruth/Labelers/BoundingBoxLabeler.cs
StevenBorkman: These are just Unity UI classes that I haven't put custom scripts on. So, I'm kind of doing it the Unity way, but maybe there is a better way that I don't know about? I always chalk this up to some of the ugly parts of dynamic UI scripting.

JonathanHUnity: When I look at this from an API contract perspective, it looks like the user needs to know that the parent of the panel is a canvas/other container that they can add UI to. That's a fairly weak contract. It would be nice to expose this more explicitly to make that contract more clear. It would also let us change around the hierarchy of the UI if we need to.

com.unity.perception/Runtime/GroundTruth/Labelers/CameraLabeler.cs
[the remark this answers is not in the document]
JonathanHUnity: It feels a bit dirty to reuse the users' canvas if they have one, especially since the user might just end up doing any number of things that we don't want to that container, like delete it. At best the user would end up seeing a bunch of extra objects attached to their canvas. I just tried it out, and having multiple canvases in your scene works fine. Is there a benefit to reuse the existing canvas?

com.unity.perception/Runtime/GroundTruth/Labelers/SemanticSegmentationLabeler.cs
StevenBorkman: This kind of got cleaned up by turning it into a property. Now it appears as `visualizationEnabled = supportsVisualization`.  So it is getting called automatically, but if the user does not want their labeler to visualize, they could set supportsVisualization to false

com.unity.perception/Runtime/GroundTruth/Labelers/CameraLabeler.cs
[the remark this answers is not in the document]
JonathanHUnity: One more note on this. In my head, the implementation details of any development tool should be hidden from the user whenever possible. This can be a real challenge in Unity, where user options in the project can sometimes break the implementation of your feature. (for example, what happens if the user changes their canvas to World Space?) It's not always possible to avoid this, but it's something to be conscious of.

com.unity.perception/Runtime/GroundTruth/Labelers/BoundingBoxLabeler.cs
StevenBorkman: So, are you suggesting to get rid of prefabs, and just create the genericToggle and generic slider on the fly? If so, that would allow us to get rid of resources all together. Unfortunately, I'll have to fight a little bit with rect transform layouts, which was a big pain, and the overall reason why I went the prefab direction to begin with.

JonathanHUnity: It it makes sense to do so, yes. At least what I'm saying is that users should not have to write `GameObject.Instantiate(Resources.Load<GameObject>("GenericToggle"))` . Instead, they should be able to do something like `Visualization.CreateToggle()` or even just `AddToggle()`

com.unity.perception/Runtime/GroundTruth/Labelers/CameraLabeler.cs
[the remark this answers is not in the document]
StevenBorkman: Reimplemented a lot of the way I am allocating UI objects, how the canvas operates, etc. But I agree with you about having our own canvas no matter what.

com.unity.perception/Runtime/GroundTruth/Labelers/BoundingBoxLabeler.cs
StevenBorkman: This has been fixed as a part of my UI rearchitecture

com.unity.perception/Runtime/GroundTruth/Labelers/Resources/BoundingBoxPrefab.prefab
StevenBorkman: I think we may need to spin off a task and try to get a smart way to do this. maybe using assetbundles or something like that. we could also look into using the new UI Elements, see if that makes it easier to create UIs via scripts

com.unity.perception/Runtime/GroundTruth/Labelers/Resources/KeyValuePanel.cs
StevenBorkman: Done.

com.unity.perception/Runtime/GroundTruth/Labelers/Visualization/ControlPanel.cs
```diff
@@ -0,0 +1,128 @@
...
+using UnityEngine.UI;
+
+namespace UnityEngine.Perception.GroundTruth
+{
+    /// <summary>
+    /// The control panel for labeler visualizers. This panel resides in the upper right hand corner of
+    /// the display. Any UI element can be added to the panel, but it contains helper methods to quickly
+    /// create some of the most common control panel display elements: toggles and sliders.
+    /// </summary>
+    public class ControlPanel : MonoBehaviour
```
JonathanHUnity, line 14: For completeness, let there be a way of reading the list of controls. Also, what happens when two different labelers want to use the same name for their control (like "enabled")? Unlikely, but consider a model where we don't have keys and instead just add and remove GameObjects

JonathanHUnity:
Things to test:
* HDRP
* Does everything scale well with large numbers of objects? (SynthDet, Smart Camera)
* Is data for deleted/hidden objects removed?

com.unity.perception/Runtime/GroundTruth/Labelers/BoundingBoxLabeler.cs
StevenBorkman: Fixed (as you know since you helped)

com.unity.perception/Runtime/GroundTruth/Labelers/BoundingBoxLabeler.cs
StevenBorkman: This has all been reworked

com.unity.perception/Runtime/GroundTruth/Labelers/Visualization/ControlPanel.cs
StevenBorkman: Ok, moved away from labels and added a list Getter

StevenBorkman: OK, updated with your suggestions.

JonathanHUnity: `ToList` allocates a new list and copies the objects, which is not desirable. [This page has some relevant bits](https://docs.microsoft.com/en-us/dotnet/standard/design-guidelines/guidelines-for-collections).

I'd suggest changing it to `IEnumerable<GameObject>`

StevenBorkman: Makes sense, updated.

PR #53, opened 2020-08-07
Segmentation Alpha Bug Fix

StevenBorkman:
Capped the max opacity allowed in the shader to an extremely high value (0.995) to stop a weird issue caused by alpha being set to 1.0

# Peer Review Information:

## Dev Testing:
**Tests Added**: 
<br>
<br>
**Core Scenario Tested**: 
<br>
**At Risk Areas**: 
<br>
**Notes + Expectations**:

### DIFF
--- a/com.unity.perception/Runtime/GroundTruth/Labelers/Visualization/Resources/SegRemoveBackgroundShader.shader
+++ b/com.unity.perception/Runtime/GroundTruth/Labelers/Visualization/Resources/SegRemoveBackgroundShader.shader
@@ -53,18 +53,10 @@
                 // sample the texture
                 fixed4 col = tex2D(_MainTex, i.uv);
 
-                if (abs(col.r - _RemoveColor.r) < .01f)
-                {
-                    if (abs(col.g - _RemoveColor.g) < .01f)
-                    {
-                        if (abs(col.b - _RemoveColor.b) < .01f)
-                        {
-                            col.a = _BackTransparency;
-                        }
-                    }
-                }
-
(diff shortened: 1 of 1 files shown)


com.unity.perception/Runtime/GroundTruth/Labelers/Visualization/Resources/SegRemoveBackgroundShader.shader
```diff
@@ -50,21 +50,26 @@
 
             fixed4 frag (v2f i) : SV_Target
             {
+                // Cap the max Transparency of the background so that we never
+                // reach 1.0 which causes an anomaly that the _SegmentTransparency
+                // value stops being honored.
+                if (_BackTransparency > 0.998f) _BackTransparency = 0.998f;
```
JonathanHUnity, line 56: I think the issue you're seeing is that the _BackTransparency of 1.0 is matching the alpha channel of the objects themselves (which in our labelers is always 1.0, but is not necessarily always such).

What if you change it to
```
if (abs(col.r - _RemoveColor.r) < .001f && abs(col.g - _RemoveColor.g) < .001f && ... )
   col.a = _BackTransparency;
else
   col.a = _SegmentTransparency;
```

StevenBorkman: You are right. Incorporated your bug fix.

JonathanHUnity:
Looks good to me!

PR #60, opened 2020-08-20
Fixing LabelEntryMatchCache 

JonathanHUnity:
# Peer Review Information:
Fixing bug where instance ids that do not match any IdLabelConfig entries are reported as matching the first entry, resulting in bounding boxes being drawn around objects whose labels have no matches in the IdLabelConfig.

## Dev Testing:
**Tests Added**: 
Adding tests for LabelEntryMatchCache
<br>
<br>
**Core Scenario Tested**: 
<br>
**At Risk Areas**: 
<br>
**Notes + Expectations**:

### DIFF
--- a/com.unity.perception/Runtime/GroundTruth/Labeling/IdLabelEntry.cs
+++ b/com.unity.perception/Runtime/GroundTruth/Labeling/IdLabelEntry.cs
@@ -5,7 +5,7 @@ namespace UnityEngine.Perception.GroundTruth {
     /// An entry for <see cref="IdLabelConfig"/> mapping a label to an integer id.
     /// </summary>
     [Serializable]
-    public struct IdLabelEntry : ILabelEntry
+    public struct IdLabelEntry : ILabelEntry, IEquatable<IdLabelEntry>
     {
         string ILabelEntry.label => this.label;
         /// <summary>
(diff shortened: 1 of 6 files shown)


com.unity.perception/Runtime/GroundTruth/Labeling/IdLabelEntry.cs
```diff
@@ -16,5 +16,26 @@ public struct IdLabelEntry : ILabelEntry
         /// The id to associate with the label.
         /// </summary>
         public int id;
+
+        /// <inheritdoc/>
+        public bool Equals(IdLabelEntry other)
+        {
+            return label == other.label && id == other.id;
```
mrpropellers, line 23: Doesn't need to be addressed in this PR, but can we end up with a label that has the same string as another label but a different ID, or same ID but different strings?

com.unity.perception/Tests/Runtime/GroundTruthTests/LabelEntryMatchCacheTests.cs
```diff
@@ -0,0 +1,93 @@
...
+        [UnityTest]
+        public IEnumerator TryGet_ReturnsFalse_ForNonMatchingLabel_WithOtherMatches()
+        {
+            var label = "label";
+            //only way to guarantee registration order is to run frames.
+            //We want to ensure labeledPlane is registered before labeledPlane2 so that the cache does not early out
+            var labeledPlane = TestHelper.CreateLabeledPlane(label: "foo");
+            AddTestObjectForCleanup(labeledPlane);
+            yield return null;
+            var labeledPlane2 = TestHelper.CreateLabeledPlane(label: label);
```
mrpropellers, line 71: kind of an anti-pattern here: one argument gets a variable name but the other doesn't

PR #69, opened 2020-08-27
Feature/lens distortion

idanbeck:
# Peer Review Information:
These changes introduce a Lens Distortion pass that is included in both semantic and instance segmentation flows.  This has been tested for both URP and HDRP, utilizing a camera that enabled post-processing with a Global Volume that overrides the Lens Distortion parameters.  This is unlikely to be the way we would like to do these effects in the long run, but at least for the time being allow ground truth to be correctly distorted per the parameters of the distortion applied to the RGB images.  

## Dev Testing:
**Tests Added**: 
SemanticSegmentationPass_WithLensDistortion

**Core Scenario Tested**: 
Test involves running the actual lens distortion shader on a captured instance segmentation mask and with and without "intensity" (which is what resolves into the actual distortion of the effect) and then ensuring the two calculated bounding boxes are not equivalent as well as making sure that the distorted bounding box with bigger (since barrel distortion is applied for intensity values over 0.5).  This test does differ slightly from the core-pipeline as it utilizes a Lens Distortion override for the pass - which essentially then defaults the values vs. using them from the volume, as the VolumeManager was not clearly accessible from the test framework (would have led to a different test for HDRP vs. URP as well, so this was more eloquent). 

**At Risk Areas**: 
The actual distortion is "lifted" from the respective UberPost shader / PostProcessing.cs implementation, so this is a limitation of these changes.  It was not obvious how to utilize these capabilities without cargo-coding these aspects, and this actually came in handy as the implementation was manipulated to allow for some generality for test purposes. 

**Notes + Expectations**: 

## Checklist
- [x] - Updated docs
- [x] - Updated changelog

### DIFF
--- a/com.unity.perception/Runtime/GroundTruth/GroundTruthCrossPipelinePass.cs
+++ b/com.unity.perception/Runtime/GroundTruth/GroundTruthCrossPipelinePass.cs
@@ -6,7 +6,7 @@ using UnityEngine.Rendering;
 
 namespace UnityEngine.Perception.GroundTruth
 {
-    public abstract class GroundTruthCrossPipelinePass : IGroundTruthGenerator
+    internal abstract class GroundTruthCrossPipelinePass : IGroundTruthGenerator
     {
         public Camera targetCamera;
 
--- a/com.unity.perception/Runtime/GroundTruth/GroundTruthCrossPipelinePass.cs
+++ b/com.unity.perception/Runtime/GroundTruth/GroundTruthCrossPipelinePass.cs
@@ -6,7 +6,7 @@ using UnityEngine.Rendering;
 
 namespace UnityEngine.Perception.GroundTruth
 {
-    abstract class GroundTruthCrossPipelinePass : IGroundTruthGenerator
+    public abstract class GroundTruthCrossPipelinePass : IGroundTruthGenerator
     {
         public Camera targetCamera;
 
--- a/com.unity.perception/Runtime/GroundTruth/LensDistortionCrossPipelinePass.cs
+++ b/com.unity.perception/Runtime/GroundTruth/LensDistortionCrossPipelinePass.cs
@@ -12,7 +12,7 @@ namespace UnityEngine.Perception.GroundTruth
     {
         const string k_ShaderName = "Perception/LensDistortion";
 
-        static int s_LastFrameExecuted = -1;
+        //static int s_LastFrameExecuted = -1;
 
         //Serialize the shader so that the shader asset is included in player builds when the SemanticSegmentationPass is used.
         //Currently commented out and shaders moved to Resources folder due to serialization crashes when it is enabled.
@@ -54,15 +54,16 @@ namespace UnityEngine.Perception.GroundTruth
 
                 m_distortedTexture = new RenderTexture(Screen.width, Screen.height, 0, RenderTextureFormat.ARGBFloat, RenderTextureReadWrite.Linear);
                 m_distortedTexture.enableRandomWrite = true;
+                m_distortedTexture.filterMode = FilterMode.Point;
                 m_distortedTexture.Create();
             }
         }
 
         protected override void ExecutePass(ScriptableRenderContext renderContext, CommandBuffer cmd, Camera camera, CullingResults cullingResult)
         {
-            if (s_LastFrameExecuted == Time.frameCount)
-                Debug.LogError("Lens Distortion executed twice in the same frame, this may lead to undesirable results.");
-            s_LastFrameExecuted = Time.frameCount;
+            //if (s_LastFrameExecuted == Time.frameCount)
+            //    Debug.LogError("Lens Distortion executed twice in the same frame, this may lead to undesirable results.");
+            //s_LastFrameExecuted = Time.frameCount;
 
             // Blitmayhem
             cmd.Blit(m_TargetTexture, m_distortedTexture, m_LensDistortionMaterial);
--- a/com.unity.perception/Runtime/GroundTruth/LensDistortionCrossPipelinePass.cs
+++ b/com.unity.perception/Runtime/GroundTruth/LensDistortionCrossPipelinePass.cs
@@ -70,6 +70,8 @@ namespace UnityEngine.Perception.GroundTruth
 
         protected override void ExecutePass(ScriptableRenderContext renderContext, CommandBuffer cmd, Camera camera, CullingResults cullingResult)
         {
+            // Review: This has been removed since we may apply this pass to various stages, just want to confirm I'm
+            // not missing anything before removing it
             //if (s_LastFrameExecuted == Time.frameCount)
             //    Debug.LogError("Lens Distortion executed twice in the same frame, this may lead to undesirable results.");
             //s_LastFrameExecuted = Time.frameCount;
@@ -85,21 +87,16 @@ namespace UnityEngine.Perception.GroundTruth
 
         public override void SetupMaterialProperties(MaterialPropertyBlock mpb, Renderer renderer, Labeling labeling, uint instanceId)
         {
-            // Grab the Lens Distortion from Perception Camera (collected in OnBeginCameraRendering
-            //PerceptionCamera perceptionCamera = this.targetCamera.GetComponent<PerceptionCamera>();
-            //if(perceptionCamera != null)
-            //{
-            //    m_LensDistortion = perceptionCamera.m_LensDistortion;
-            //}
+            // Grab the Lens Distortion from Perception Camera stack
+            var hdCamera = HDCamera.GetOrCreate(targetCamera);
+            var stack = hdCamera.volumeStack;
+            var lensDistortion = stack.GetComponent<LensDistortion>();
+            Debug.Log("lens distortion intensity: " + lensDistortion.intensity.value);
 
             // This code is lifted from the SetupLensDistortion() function in
             // https://github.com/Unity-Technologies/Graphics/blob/257b08bba6c11de0f894e42e811124247a522d3c/com.unity.render-pipelines.universal/Runtime/Passes/PostProcessPass.cs
             // This is in UnityEngine.Rendering.Universal.Internal.PostProcessPass::SetupLensDistortion so it's
             // unclear how to re-use this code
-
-            var stack = VolumeManager.instance.stack;
-            var lensDistortion = stack.GetComponent<LensDistortion>();
-            //Debug.Log("lens distortion intensity: " + lensDistortion.intensity.value);
 
             float amount = 1.6f * Mathf.Max(Mathf.Abs(lensDistortion.intensity.value * 100.0f), 1.0f);
             float theta = Mathf.Deg2Rad * Mathf.Min(160f, amount);
--- a/com.unity.perception/Runtime/GroundTruth/LensDistortionCrossPipelinePass.cs
+++ b/com.unity.perception/Runtime/GroundTruth/LensDistortionCrossPipelinePass.cs
@@ -8,64 +8,75 @@ namespace UnityEngine.Perception.GroundTruth
     /// Custom Pass which renders labeled images where each object labeled with a Labeling component is drawn with the
     /// value specified by the given LabelingConfiguration.
     /// </summary>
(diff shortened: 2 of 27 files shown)


com.unity.perception/Runtime/GroundTruth/GroundTruthCrossPipelinePass.cs
```diff
@@ -6,7 +6,7 @@
 
 namespace UnityEngine.Perception.GroundTruth
 {
-    abstract class GroundTruthCrossPipelinePass : IGroundTruthGenerator
+    public abstract class GroundTruthCrossPipelinePass : IGroundTruthGenerator
```
JonathanHUnity, line 9: This needs to be documented if it is to be made public. Probably easier to just make it internal again though.

com.unity.perception/Runtime/GroundTruth/LensDistortionCrossPipelinePass.cs
```diff
@@ -0,0 +1,181 @@
...
+{
+    /// <summary>
+    /// Custom Pass which renders labeled images where each object labeled with a Labeling component is drawn with the
+    /// value specified by the given LabelingConfiguration.
+    /// </summary>
+    public class LensDistortionCrossPipelinePass : GroundTruthCrossPipelinePass
+    {
+        const string k_ShaderName = "Perception/LensDistortion";
+
+        //static int s_LastFrameExecuted = -1;
```
JonathanHUnity, line 22: some comments here to clean up

com.unity.perception/Runtime/GroundTruth/LensDistortionCrossPipelinePass.cs
```diff
@@ -0,0 +1,181 @@
...
+    using UnityEngine.Rendering.Universal;
+#endif
+
+namespace UnityEngine.Perception.GroundTruth
+{
+    /// <summary>
+    /// Custom Pass which renders labeled images where each object labeled with a Labeling component is drawn with the
+    /// value specified by the given LabelingConfiguration.
+    /// </summary>
+    public class LensDistortionCrossPipelinePass : GroundTruthCrossPipelinePass
```
JonathanHUnity, line 18: out of date docs

com.unity.perception/Runtime/GroundTruth/LensDistortionCrossPipelinePass.cs
```diff
@@ -0,0 +1,181 @@
...
+        //Serialize the shader so that the shader asset is included in player builds when the SemanticSegmentationPass is used.
+        //Currently commented out and shaders moved to Resources folder due to serialization crashes when it is enabled.
+        //See https://fogbugz.unity3d.com/f/cases/1187378/
+        //[SerializeField]
+
+        // Lens Distortion Shader
+        private Shader m_LensDistortionShader;
+        private Material m_LensDistortionMaterial;
+
+        private bool fInitialized = false;
```
JonathanHUnity, line 33: Per our code conventions, private should be omitted where possible and private fields should start with m_. I'd recommend following this guide to get tooling support http://qatestrail.hq.unity3d.com/index.php?/runs/view/25567&group_by=cases:section_id&group_order=asc

com.unity.perception/Runtime/GroundTruth/LensDistortionCrossPipelinePass.cs
JonathanHUnity: Also, the public API here will need to be documented

com.unity.perception/Runtime/GroundTruth/LensDistortionCrossPipelinePass.cs
```diff
@@ -0,0 +1,181 @@
...
+#endif
+
+            fInitialized = true;
+        }
+
+        protected override void ExecutePass(ScriptableRenderContext renderContext, CommandBuffer cmd, Camera camera, CullingResults cullingResult)
+        {
+            // Review: This has been removed since we may apply this pass to various stages, just want to confirm I'm
+            // not missing anything before removing it
+            //if (s_LastFrameExecuted == Time.frameCount)
```
JonathanHUnity, line 98: +1 on removing. I don't think there is anything wrong with running this multiple times in the same frame.

com.unity.perception/Runtime/GroundTruth/PerceptionCamera.cs
```diff
@@ -467,6 +467,25 @@ public bool RemoveLabeler(CameraLabeler cameraLabeler)
             return false;
         }
 
+        /// <summary>
+        /// Retrieves the given <see cref="CameraLabeler"/> from the list of labelers under this PerceptionCamera, if it
+        /// is in the list.  Returns null if not found
+        /// </summary>
+        /// <param name="cameraLabeler"></param>
+        /// <returns></returns>
+        public CameraLabeler GetLabeler<T>()
```
JonathanHUnity, line 476: Is this used anywhere? It's not bad to add, but I didn't include it before because it is valid to have multiple CameraLabelers of the same type.

com.unity.perception/Runtime/GroundTruth/PerceptionCamera_InstanceSegmentation.cs
```diff
@@ -29,12 +29,26 @@ partial class PerceptionCamera
...
 
+    #if HDRP_PRESENT
+        InstanceSegmentationPass m_InstanceSegmentationPass;
+        LensDistortionPass m_LensDistortionPass;
+    #elif URP_PRESENT
+        InstanceSegmentationUrpPass m_InstanceSegmentationPass;
+        LensDistortionUrpPass m_LensDistortionPass;
+    #endif
+
+        public void OverrideLensDistortionIntensity(float intensity)
```
JonathanHUnity, line 40: If this is only for testing, could the tests use Volume and set the lens distortion directly? Or, could this be made internal?

com.unity.perception/Runtime/GroundTruth/PerceptionCamera_InstanceSegmentation.cs
```diff
@@ -43,17 +57,31 @@ void SetupInstanceSegmentation()
...
-#endif
-#if URP_PRESENT
-            AddScriptableRenderPass(new InstanceSegmentationUrpPass(myCamera, m_InstanceSegmentationTexture));
+            m_InstanceSegmentationPass.EnsureInit();
+            customPassVolume.customPasses.Add(m_InstanceSegmentationPass);
+
+            // TODO: Note - the implementation here differs substantially from how things are done in SemanticSegmentationLalber
+            // Also, the naming convention doesn't line up, shouldn't instance segmentation be a labeller?  At least in
+            // architecture
+            m_LensDistortionPass = new LensDistortionPass(GetComponent<Camera>(), m_InstanceSegmentationTexture)
```
JonathanHUnity, line 72: Which part of the implementation is different?
We decided to keep instance segmentation itself out of a labeler for now because we didn't want to have to deal with dependencies between labelers. Also, there is no configuration of the instance segmentation logic, so there is no point in having more than one. At some point we'll have an instance segmentation labeler though to save the images to disk, but perhaps not before the new architecture :)

com.unity.perception/Runtime/GroundTruth/Resources/LensDistortion.shader
```diff
@@ -0,0 +1,128 @@
...
+    //{
+    //    [PerObjectData] _SegmentationId("Segmentation ID", int) = 0
+    //}
+
+    Properties
+    {
+        _MainTex ("Texture", 2D) = "white" {}
+    }
+
+    /*
```
JonathanHUnity, line 13: remove comments?

com.unity.perception/Runtime/GroundTruth/SemanticSegmentationCrossPipelinePass.cs
```diff
@@ -33,11 +33,20 @@ public override void Setup()
...
-            shaderVariantCollection.WarmUp();
+
+            if (shaderVariantCollection != null)
+            {
+                shaderVariantCollection.Add(new ShaderVariantCollection.ShaderVariant(m_ClassLabelingShader, PassType.ScriptableRenderPipeline));
+            }
 
             m_OverrideMaterial = new Material(m_ClassLabelingShader);
+
+            if (shaderVariantCollection != null)
```
JonathanHUnity, line 46: Will it ever be null?

com.unity.perception/Runtime/GroundTruth/PerceptionCamera.cs
idanbeck: This was part of a dead-end in trying to get this all to work in the test, but ultimately I went with a test hook approach due to it's being a more direct way to get a given value into the shader and as such a more reliable test.  However, figured this might be useful functionality.  One approach would be to return a container instead of a single CameraLabeler - or do a string lookup to specify?  Alternatively, could just get rid of this - it's not doing anything in my changes at this point.

JonathanHUnity: I don't have a strong opinion either way. GameObject.GetComponent<T> has the same behavior, even though there can technically be more than one of the same type of component attached.

com.unity.perception/Runtime/GroundTruth/LensDistortionCrossPipelinePass.cs
idanbeck: Apart from updating this in the comment, where would I need to update the documentation?  Or is this more a comment that this text should be documenting the actual functionality of the pass (which is what I've addressed in an incoming change).

com.unity.perception/Runtime/GroundTruth/PerceptionCamera_InstanceSegmentation.cs
idanbeck: Yeah - I'll back off this point for now, and defer to hashing this out in arch!

com.unity.perception/Runtime/GroundTruth/SemanticSegmentationCrossPipelinePass.cs
idanbeck: I just have trouble seeing stuff like this in engine code - as in, not much error handling.  I know the underlying engine is generally something we can depend on, but I'm used to always adding debug null checks which get removed in release/production

com.unity.perception/Runtime/GroundTruth/LensDistortionCrossPipelinePass.cs
JonathanHUnity: Just that the text should reflect the feature. I don't necessarily think we need to add anything to the manual since this feature "just works" with no new UI or user control.

com.unity.perception/Runtime/GroundTruth/SemanticSegmentationCrossPipelinePass.cs
JonathanHUnity: In general we have the benefit of using very established features that generally follow their public contracts :)

com.unity.perception/Runtime/GroundTruth/LensDistortionCrossPipelinePass.cs
idanbeck: Ok great - I did add change the text per that.  Doesn't look like github updated it here however.

com.unity.perception/Runtime/GroundTruth/SemanticSegmentationCrossPipelinePass.cs
idanbeck: Makes sense - this was updated in a prior change, but github has not updated it here.

com.unity.perception/Runtime/GroundTruth/PerceptionCamera.cs
idanbeck: I'll remove it for the time being - unclear what it'll be used for regardless.

idanbeck: Not a huge fan of checking in code that isn't being used for anything and my introduce an unknown bug as a result if someone down the line actually does try to use the functionality to find that it wasn't built correctly.

PR #78, opened 2020-09-11
Fixes math in SimulationState SequenceTimeOfNextCapture

mrpropellers:
# Peer Review Information:
SequenceTimeOfNextCapture was reporting time with respect to two
different frames depending on context: either it would report absolute
time since the start of the sequence, if the first capture had not been
completed, or it would report the amount of time left until the next
capture, relative to the current time.

Removed logic which reported relative time, which seems like it was not
part of any code path, or was being corrected for elsewhere in
SimulationState's update loop. Removed some of the error correction in
the update loop and added stricter expectations for the management of
the timing code.

## Dev Testing:
Added unit tests to ensure function reports time we'd expect regardless
of context.

**At Risk Areas**: 
Handling of the SensorData struct is pretty loosey-goosey in SimulationState, and there are some assumptions being made about update ordering that are easy to violate. Small modifications to how the timing values are computed cause cascading failures across pretty much every other ground truth related component.

**Notes + Expectations**: 
Created follow-up task AISV-845 for making the timing code less fragile.

### DIFF
--- a/com.unity.perception/Runtime/GroundTruth/SimulationState.cs
+++ b/com.unity.perception/Runtime/GroundTruth/SimulationState.cs
@@ -4,6 +4,7 @@ using System.IO;
 using System.Linq;
 using Newtonsoft.Json.Linq;
 using Unity.Collections;
+using Unity.Mathematics;
 using Unity.Simulation;
 using UnityEngine;
 using UnityEngine.Profiling;
@@ -378,10 +379,10 @@ namespace UnityEngine.Perception.GroundTruth
 
         float SequenceTimeOfNextCapture(SensorData sensorData)
         {
+            // If the first capture hasn't happened yet, sequenceTimeNextCapture field won't be valid
             if (sensorData.firstCaptureTime >= UnscaledSequenceTime)
                 return sensorData.firstCaptureTime;
-
-            return sensorData.period - (UnscaledSequenceTime - sensorData.firstCaptureTime) % sensorData.period;
+            return sensorData.sequenceTimeNextCapture;
         }
 
         public bool Contains(Guid id) => m_Ids.Contains(id);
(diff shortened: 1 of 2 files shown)


com.unity.perception/Runtime/GroundTruth/SimulationState.cs
```diff
@@ -4,6 +4,7 @@
 using System.Linq;
 using Newtonsoft.Json.Linq;
 using Unity.Collections;
+using Unity.Mathematics;
```
mrpropellers, line 7: Dangling include; will remove.

com.unity.perception/Tests/Runtime/GroundTruthTests/DatasetCaptureSensorSchedulingTests.cs
```diff
@@ -7,16 +9,78 @@
...
+    // due to protection level - use only when testing protected logic is critical
+    class SimulationStateTestHelper
+    {
+        SimulationState m_State => DatasetCapture.SimulationState;
+        Dictionary<SensorHandle, SimulationState.SensorData> m_SensorsReference;
+        MethodInfo m_SequenceTimeOfNextCaptureMethod;
+
+        internal SimulationStateTestHelper()
+        {
+            var bindingFlags = BindingFlags.NonPublic | BindingFlags.Instance;
```
JonathanHUnity, line 22: Instead of reflection, it's cleaner to mark the members you need `internal`. Then the API will be available in your tests.

JonathanHUnity (changes requested):
Just one request regarding the use of reflection.

com.unity.perception/Tests/Runtime/GroundTruthTests/DatasetCaptureSensorSchedulingTests.cs
mrpropellers: I think it depends on your definition of "cleaner"  - less code? Sure.  But, changing protection levels to enable unit tests also feels... pretty dirty.  In a nominal case we'd want to be testing this math at the API level, rather than turning the math itself into part of the internal API. Given that we've decided this is worth testing, even as private method, though, I'd rather hack the unit test than hack the interface.

JonathanHUnity: Testing internal state is unfortunately fairly common, and not always is it possible to use reflection to do the testing you need. For example, you are already using the internal type SimulationState.SensorData in this API. It was marked public (in an internal class) originally so that it could be used in tests, and to do otherwise would be a much bigger pain than just calling a method or assigning a field. While it's nice to not change the internal API of a class, it doesn't scale well in general and increases maintenance cost of tests.

This isn't a huge deal at the end of the day so I'll mark this approved and you can do what you think is right.

JonathanHUnity (approved):

com.unity.perception/Tests/Runtime/GroundTruthTests/DatasetCaptureSensorSchedulingTests.cs
mrpropellers: I'm going to push this, but I have a feeling we're going to have this conversation again, so I'm also going to list my justifications, as there are quite a few red flags in your line of thinking:
1. "it's common" is not justification for employing bad practices
2. SimulationState.SensorData being public is tech debt. Using existing tech debt to justify adding more tech debt leads to an unmaintainable codebase
3. I did not make the claim that this scales well, and in fact it's not meant to. Testing at this level of granularity implicitly does not scale well and is in general bad practice. Because timing code tends to be critical, I'm ok with adding this test,  but this is an exception to the rule. It _should_ be hard to scale up testing in ways that do not follow good practices, and we _should not_ use one-off exceptions like this to set precedent for future bad decisions.

PR #82, opened 2020-10-03
Aisv565 bb3d - Bounding box 3D perception changes

StevenBorkman:
# Peer Review Information:

## Dev Testing:
**Tests Added**: 

**Core Scenario Tested**: 

**At Risk Areas**: 

**Notes + Expectations**: 

## Checklist
- [x] - Updated docs
- [x] - Updated changelog


com.unity.perception/Runtime/GroundTruth/Labelers/BoundingBox3DLabeler.cs
```diff
@@ -0,0 +1,256 @@
...
+                instance_id = instanceId,
+                translation = new[] { center.x, center.y, center.z },
+                size = new[] { extents.x * 2, extents.y * 2, extents.z * 2 },
+                rotation = new[] { rotation.x, rotation.y, rotation.z, rotation.w },
+                velocity = null,
+                acceleration = null
+            };
+        }
+
+        static Vector3[] GetBoxCorners(Bounds bounds, Quaternion rotation)
```
idanbeck, line 141: This is a great solution

com.unity.perception/Runtime/GroundTruth/Labelers/GroundTruthUpdateSystem.cs
```diff
@@ -0,0 +1,48 @@
+using System.Collections.Generic;
+using Unity.Entities;
+using UnityEngine;
+
+namespace UnityEngine.Perception.GroundTruth
```
idanbeck, line 5: So one comment is with regards to how this is done - since instead of an update system, it could be possible to instead create the respective bounding boxes each time you render the scene.  In other words, you're not registering an object with an update system but rather running through and calculating all of the bounding boxes before rendering the scene. 

I'm split on this, as it's a marginal difference and largely has to do with the frequency of getting the OnUpdate in the GroundTruthUpdateSystem object.

com.unity.perception/Runtime/GroundTruth/PerceptionCamera.cs
```diff
@@ -312,12 +313,25 @@ void OnValidate()
                 m_Labelers = new List<CameraLabeler>();
         }
 
+        // Convert the Unity 4x4 projection matrix to a 3x3 matrix
+        // ReSharper disable once InconsistentNaming
+        static float3x3 ToProjectionMatrix3x3(Matrix4x4 inMatrix)
```
idanbeck, line 318: An alternative to this is taking the outMatrix = transpose(inverse(inMatrix)) 

Less performant, but I think more correct in terms of maintaining the dimensionality of the transform.  Also will correctly handle scaling in case det(matrix) != 1  (a bit of hand waiving here, would confirm this before actually doing it!)

idanbeck (approved):
This looks great - but as mentioned in the review, I think there could be some conversation about whether or not to actually add this object registration/update loop vs. just determine the bounding boxes as a one-shot process prior to rendering.  The end result is about the same I think, the assumption being that the OnUpdate is hit sequentially to the rendering of the perception camera.

com.unity.perception/Runtime/GroundTruth/PerceptionCamera.cs
StevenBorkman: So, my main issue that I have here, is that the dataset insights code just wants to use a 3x3 matrix. From what I could make sense of, it looks like they were just using it for rotation (and I guess scaling) but the position was ignored. Also, I don't think that they had any use for sheering. Anyway, I'm not really sure what I need to do to go from a 4x4 to a 3x3 but it appeared that in a 4x4 homogenous matrix, the upper 3x3 correlate directly with a 3x3 affine transformation. But I really don't know

com.unity.perception/Runtime/GroundTruth/Labelers/GroundTruthUpdateSystem.cs
JonathanHUnity: +1 on this. Could you move the query for labeled objects ( `m_QueryBuilder.ForEach` ) into the BoundingBox3DLabeler's OnBeginUpdate method?

com.unity.perception/Runtime/GroundTruth/Labelers/BoundingBox3DLabeler.cs
```diff
@@ -0,0 +1,256 @@
...
+                    BoundingBoxComputed?.Invoke(m_CurrentFrame, converted);
+                    m_BoundingBoxValues.Add(converted);
+                }
+            }
+        }
+
+        /// <inheritdoc/>
+        public void OnEndUpdate()
+        {
+            perceptionCamera.SensorHandle.ReportAnnotationAsync(m_AnnotationDefinition).ReportValues(m_BoundingBoxValues);
```
JonathanHUnity, line 253: I don't see anything for only reporting objects that are visible in the frame, which was part of the v1 design before. Am I missing it?

JonathanHUnity (changes requested):
Couple issues here, but code is great overall.

com.unity.perception/Runtime/GroundTruth/PerceptionCamera.cs
idanbeck: I would argue that they should be doing it the correct linear algebra way too in that case :) 

Stripping the matrix down doesn't compensate for scale properly since it's possible for mat_4_4 to be non 1 for example and this would then in effect scale the rest of the matrix down, which is why you want to do that transpose-inverse thing. 

One idea is to just maintain it as 4x4 Unity side all the way through and then export it as 3x3, although I'd argue better to export as 4x4 and let dataset insights convert it. 

This could also be chalked up to convention as well, so feel free to ignore all of this as well - functionally it really doesn't make much of a difference

com.unity.perception/Runtime/GroundTruth/Labelers/BoundingBox3DLabeler.cs
StevenBorkman: No, you are correct, I completely forgot about that. I will implement that immediately.

StevenBorkman: Ok, this has been fixed

com.unity.perception/Runtime/GroundTruth/Labelers/GroundTruthUpdateSystem.cs
StevenBorkman: @JonathanHUnity So are you suggesting that instead of having the GroundTruthUpdateSystem iterate through each entity in the query, have each labeler execute the query itself? Would that buy us anything? I feel like it definitely would add complexity to any future plugin development, each new labeler would have to realize that they had to perform the same step.

StevenBorkman: I think what @idanbeck is suggesting is really interesting, but it would add quite a bit of rework when it might be best to get this in there for right now. Especially with a re-architecture looming in the future.

JonathanHUnity: Yes, essentially inlining the code for doing the only query. It doesn't feel great to not be calling an explicit API for this, but in DOTS, the data is the API. It's ok to have to know what components to query for. Once you get into this site, it actually simplifies the code quite a bit, especially around timing since it reduces callbacks and events.

com.unity.perception/Runtime/GroundTruth/Labelers/BoundingBox3DLabeler.cs
```diff
@@ -0,0 +1,278 @@
...
+        }
+
+        static BoxData ConvertToBoxData(IdLabelEntry label, uint instanceId, Vector3 center, Vector3 extents, Quaternion rotation)
+        {
+            return new BoxData
+            {
+                label_id = label.id,
+                label_name = label.label,
+                instance_id = instanceId,
+                translation = new[] { center.x, center.y, center.z },
```
JonathanHUnity, line 137: These should be able to be Vector3 or Quaternions. Try to avoid allocating arrays, as the GC is slow.

StevenBorkman: I'm doing this because Vector3 and Quaternions are not serializable. I'm not sure if we want to add a way to work around this.

com.unity.perception/Runtime/GroundTruth/Labelers/GroundTruthUpdateSystem.cs
StevenBorkman: Ok, this has been fixed

com.unity.perception/Runtime/GroundTruth/Labelers/BoundingBox3DLabeler.cs
```diff
@@ -0,0 +1,277 @@
...
+            perceptionCamera.SensorHandle.ReportAnnotationAsync(m_AnnotationDefinition).ReportValues(m_BoundingBoxValues);
+        }
+
+        bool IsInCameraView(GameObject target)
+        {
+            // Need to check against all of the meshes in the labeled object, if at least one is visible, then
+            // the object will be considered in the camera's view. The built in Unity geometry methods are optimized to
+            // work with AABB bounds, so we will use those from the mesh renderer, as opposed to the non-axis aligned
+            // bounding boxes retrieved from the mesh filter. There is a chance that this method can report false
+            // positives, but it is (at least right now) probably worth the trade off for performance
```
JonathanHUnity, line 181: There appears to be a fairly high chance of false positives here, especially when objects are occluded. It's a bit more work, but this would be addressed if we use the RenderedObjectInfo to filter the set of bounding boxes before reporting them.

com.unity.perception/Runtime/GroundTruth/Labelers/BoundingBox3DLabeler.cs
JonathanHUnity: Hmm, it looks like AnnotationAsync.ReportValues is still using JArray.FromObject() instead of JArrayFromArray. I tried serializing a struct with Vector3 and Quaternion to Json using `DatasetJsonUtility.ToJToken(hasVec);` and it worked fine, so with a little bit of rework of AnnotationAsync.ReportValues you should be able to make the switch.

com.unity.perception/Runtime/GroundTruth/Labelers/BoundingBox3DLabeler.cs
StevenBorkman: fixed.

com.unity.perception/Runtime/GroundTruth/Labelers/BoundingBox3DLabeler.cs
StevenBorkman: This has been fixed, I found a pretty cool way that we could directly support reading and writing Vector3 and Quaternions to json

com.unity.perception/Runtime/GroundTruth/SimulationState.cs
```diff
@@ -615,10 +622,61 @@ public AsyncAnnotation ReportAnnotationAsync(AnnotationDefinition annotationDefi
             ReportAsyncAnnotationResult<T>(asyncAnnotation, filename, jArray);
         }
 
-        public void ReportAsyncAnnotationResult<T>(AsyncAnnotation asyncAnnotation, string filename = null, IEnumerable<T> values = null)
+        [SuppressMessage("ReSharper", "PossibleInvalidOperationException")]
+        public class QuaternionConverter : JsonConverter<Quaternion>
         {
-            var jArray = values == null ? null : JArray.FromObject(values);
+            public override void WriteJson(JsonWriter writer, Quaternion quaternion, JsonSerializer serializer)
```
JonathanHUnity, line 628: That's an interesting approach. We have some existing code for these types which is pretty similar in DatasetJsonUtility. Give it a quick look to see if it's worth having both versions.

JonathanHUnity (approved):
Great! Just look into DatasetJsonUtility before pushing.

PR #84, opened 2020-10-06
Aisv756 obj count bug

StevenBorkman:
# Peer Review Information:

## Dev Testing:
**Tests Added**: 

**Core Scenario Tested**: 

**At Risk Areas**: 

**Notes + Expectations**: 

## Checklist
- [x] - Updated docs
- [x] - Updated changelog

### DIFF
--- a/com.unity.perception/Runtime/GroundTruth/Labelers/Visualization/HUDPanel.cs
+++ b/com.unity.perception/Runtime/GroundTruth/Labelers/Visualization/HUDPanel.cs
@@ -2,6 +2,7 @@ using System;
 using System.Collections;
 using System.Collections.Generic;
 using System.Linq;
+using System.Text;
 using UnityEngine;
 using UnityEngine.UI;
 
@@ -24,7 +25,7 @@ namespace UnityEngine.Perception.GroundTruth
         const int k_YPadding = 10;
         const int k_BoxWidth = 200;
         const int k_YLineSpacing = 4;
-
+        const int k_MaxKeyLength = 20;
 
         /// <summary>
         /// The number of labelers currently displaying real-time information on the visualization HUD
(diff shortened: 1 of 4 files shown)


com.unity.perception/Runtime/GroundTruth/Labelers/Visualization/HUDPanel.cs
```diff
@@ -110,7 +113,10 @@ internal void OnDrawGUI()
                 {
                     GUILayout.BeginHorizontal();
                     GUILayout.Space(5);
-                    GUILayout.Label(entry.Key);
+                    var k = new StringBuilder(entry.Key.Substring(0, Math.Min(entry.Key.Length, k_MaxKeyLength)));
+                    if (k.Length != entry.Key.Length)
```
JonathanHUnity, line 117: Minor issue - this approach will potentially turn "A long string" into "A long strin..." - increasing the string length instead of shortening it.

JonathanHUnity (approved):
Good. Does this also fix ground truth when stepping frame by frame?

com.unity.perception/Runtime/GroundTruth/Labelers/Visualization/HUDPanel.cs
StevenBorkman: Yeah, I guess it could be handled a little better, where we remove 3 letters off the back if shortening. The main thing I was trying to do was make sure that we didn't go pass the buffer, so that is taking into account the size of the ellipsis.

StevenBorkman:
> Good. Does this also fix ground truth when stepping frame by frame?

Unfortunately no. I'm thinking (although not 100% sure at this time) that that may be an issue with capture. I have verified that once the game is paused, that no rgb images are written to disk when frame by frame incrementing occurs.

PR #85, opened 2020-10-06
Performance testing

StevenBorkman:
# Peer Review Information:
Added performance tests along with appropriate yamato to perform nightly tests and report results to https://observer.cds.internal.unity3d.com/project

## Dev Testing:
**Tests Added**: 

A slew of performance tests

**Core Scenario Tested**: 

**At Risk Areas**: 

**Notes + Expectations**: 

## Checklist
- [x] - Updated docs
- [x] - Updated changelog


JonathanHUnity:
Yay!
image
Can you explain the data we are seeing in Observer? The numbers appear to be flipping between ~10-20ms and ~50-120ms.

com.unity.perception/Tests/Performance/PerformanceTester.cs
```diff
@@ -0,0 +1,139 @@
...
+        [UnityTest, Performance]
+        public IEnumerator ExecuteTest()
+        {
+            yield return Measure.Frames()
+                .WarmupCount(10)
+                .MeasurementCount(30)
+                .Run();
+
+            // Allow all file writes to complete
+            yield return new WaitForSeconds(5);
```
JonathanHUnity, line 136: DatasetCapture.ResetSimulaton() will take care of this

JonathanHUnity: If it doesn't then we should look into it later :)

JonathanHUnity (approved):
Just a minor point. Otherwise looks good :)

com.unity.perception/Tests/Performance/PerformanceTester.cs
StevenBorkman: Ok, I'll try that out. I actually didn't know about that call until recently, I had to put it in to fix some test errors.

StevenBorkman: It may be something that we look into again later. When I removed it, the standard deviation and max execution time rose quite a bit when running with a series of tests executing one after another.

PR #100, opened 2020-10-23
Fixed erratic behaviour were labeled objects that did not match any tracked label were still detected as foreground objects

mkamalza:
# Peer Review Information:

## Dev Testing:
**Tests Added**: 

**Core Scenario Tested**: 

**At Risk Areas**: 

**Notes + Expectations**: 

## Checklist
- [ ] - Updated docs
- [ ] - Updated changelog


JonathanHUnity (changes requested):
Could you add a test for this in LabelEntrymatchCacheTests? I'm a bit surprised this wasn't covered already, as there seem to be some related tests.

mkamalza:
I added a test. The reason the bug happens is if you spawn objects on each iteration, at the start of each iteration, there is no labeled object in the scene. So the instanceId counter resets to 1. Now new objects that are spawned for the new iteration have not been assigned instance ids yet, so they get new ids starting from 1. This is while the label entry match caches of all label configs still contain stale information about previous instance ids.

This is currently manifested by the use of iterations, but any time all labeled objects are removed from the scene, this happens. (as shown in the new test)

PR #103, opened 2020-10-29
Terrain support

JonathanHUnity:
# Peer Review Information:
Adds support for labeling terrains, with the following caveats:

* The layers on the terrain cannot be distinguished in the segmentation image
* Trees rendered using Terrain will properly occlude but will not be able to be labeled (see screenshot below)

This PR also adds terrains to the test projects SampleScene and updates them to 2019.4.12f1

## Dev Testing:
**Tests Added**: 
Added tests for terrain labeling

**Core Scenario Tested**: 
Tested on HDRP and URP

**At Risk Areas**: 

**Notes + Expectations**: 

## Checklist
- [x] - Updated docs
- [x] - Updated changelog


mkamalza (approved):
Looks good. Couple of small things:
- The "Terrain" gameobject in the URP sample scene does not have a "Terrain" label attached. HDRP is good.
- The "Directional Light" gameobject in the URP sample scene has a missing script on it. If this is not needed we should remove it.

PR #104, opened 2020-10-29
USim scenario iteration offset fix

sleal-unity:
# Peer Review Information:
This PR patches a bug in the USimScenario class that resulted in the scenario iteration incorrectly being offset before the Unity Simulation app-params had a chance to deserialize.

## Checklist
- [X] - Updated docs
- [X] - Updated changelog


mkamalza (approved):
Tested it on USim, can confirm the outputs are now different for different instances.

PR #106, opened 2020-11-03
Fixed Unity Simulation nodes generating an empty first image

sleal-unity:
# Peer Review Information:
In the unity editor and local players, the perception package cannot output the first frame of the simulation, so the randomization scenario class will skip the first frame. This is not the case on Unity Simulation. This PR introduces a check that will allow randomized perception projects to not skip the first frame if running on Unity Simulation.

## Checklist
- [X] - Updated changelog


mkamalza (approved):
Tested on USim. First images are no longer blank, and the correct number of images are produced per instance.

PR #115, opened 2020-11-12
Fixing notebook instructions

JonathanHUnity:
# Peer Review Information:
Notebook instructions in tutorial suggests using the Dataset<GUID> directory as the "data" directory, but this leads to errors in the latest datasetinsights package and I've never known this to be supported. This PR switches it to just use the dataset root as the "data" directory.

## Dev Testing:
**Tests Added**: 

**Core Scenario Tested**: 

**At Risk Areas**: 

**Notes + Expectations**: 

## Checklist
- [x] - Updated docs
- [ ] - Updated changelog


mkamalza (approved):
Thank you Jon. Curiously, it actually did work the way I had described here at the time. But being able to omit the guid is definitely much more convenient.

PR #136, opened 2020-12-11
Analytics

JonathanHUnity:
# Peer Review Information:
Adds editor analytics to the Run in Unity Simulation window. Also adds a Scenario to the sample scene.

## Dev Testing:
**Tests Added**: 
None. Not sure if it works yet :)

**Core Scenario Tested**: 

**At Risk Areas**: 

**Notes + Expectations**: 

## Checklist
- [ ] - Updated docs
- [ ] - Updated changelog


TestProjects/PerceptionURP/Assets/Scenes/SampleScene.unity
```diff
@@ -98,7 +98,7 @@ LightmapSettings:
     m_TrainingDataDestination: TrainingData
     m_LightProbeSampleCountMultiplier: 4
   m_LightingDataAsset: {fileID: 0}
-  m_UseShadowmask: 0
+  m_UseShadowmask: 1
```
mkamalza, line 101: Not related to this PR, but I noticed the URP test project has baked GI and lighting auto-generate turned on. Should we perhaps turn these off?

com.unity.perception/Editor/Randomization/Editors/PerceptionEditorAnalytics.cs
```diff
@@ -0,0 +1,99 @@
...
+        }
+
+        enum RunStatus
+        {
+            Started,
+            Failed,
+            Succeeded
+        }
+
+        struct RunInUnitySimulationData
```
mkamalza, line 37: Might be nice to add the "frames per iteration" value to this list.

com.unity.perception/Editor/Randomization/Editors/PerceptionEditorAnalytics.cs
```diff
@@ -0,0 +1,99 @@
...
+namespace UnityEngine.Perception.Randomization.Editor
+{
+    static class PerceptionEditorAnalytics
+    {
+        static int k_MaxItems = 100;
+        static int k_MaxEventsPerHour = 100;
+        const string k_VendorKey = "unity.perception";
+        const string k_RunInUnitySimulationName = "runinunitysimulation";
+
+        static bool k_IsRegistered = false;
```
mkamalza, line 15: My Rider seems to think this line should be:
`static bool s_IsRegistered;`

mkamalza (approved):

com.unity.perception/Editor/Randomization/Editors/PerceptionEditorAnalytics.cs
JonathanHUnity: Good catch. Addressed.

com.unity.perception/Editor/Randomization/Editors/PerceptionEditorAnalytics.cs
JonathanHUnity: Hmm, that would be nice but unfortunately would be a big change, since we would also need to update the db. Going to leave this out for now. I hope that at some point we will be able to grab the app params from people's runs directly from USim to get lots of metrics.

PR #142, opened 2020-12-22
Replaced individual sampler seeds with one global seed

sleal-unity:
# Peer Review Information:
This PR updates samplers to utilize a global random state provided through the scenario to generate random values instead of each sampler mutating their own unique random state.

## Dev Testing:
**Tests Changed**: 
ScenarioTests and sampler tests were adjusted to track the active scenario's random state instead of individual sampler states

## Checklist
- [X] - Updated docs
- [X] - Updated changelog


com.unity.perception/Runtime/Randomization/Samplers/SamplerTypes/NormalSampler.cs
```diff
@@ -87,8 +50,9 @@ public void IterateState(int offsetIndex)
         /// <returns>The generated sample</returns>
         public float Sample()
         {
+            var rng = new Unity.Mathematics.Random(ScenarioBase.activeScenario.NextRandomSeed());
```
StevenBorkman, line 53: 'NextRandomSeed()' makes this method sound like it is doing something different than what it is doing. I suggest renaming this method so that more closely reflects what is really happening and perhaps, updated surrounding documentation

com.unity.perception/Runtime/Randomization/Samplers/SamplerUtility.cs
```diff
@@ -42,7 +46,7 @@ public static uint GenerateRandomSeed()
         /// <param name="x">Unsigned integer to hash</param>
         /// <returns>The calculated hash value</returns>
         [MethodImpl(MethodImplOptions.AggressiveInlining)]
-        static uint Hash32(uint x) {
+        public static uint Hash32(uint x) {
```
StevenBorkman, line 49: open '{' are supposed to be on a new line

com.unity.perception/Runtime/Randomization/Samplers/SamplerUtility.cs
```diff
@@ -55,7 +59,7 @@ public static uint GenerateRandomSeed()
         /// <param name="x">64-bit value to hash</param>
         /// <returns>The calculated hash value</returns>
         [MethodImpl(MethodImplOptions.AggressiveInlining)]
-        static ulong Hash64(ulong x) {
+        public static ulong Hash64(ulong x) {
```
StevenBorkman, line 62: open '{' are supposed to be on a new line

com.unity.perception/Runtime/Randomization/Scenarios/ScenarioBase.cs
```diff
@@ -324,28 +335,13 @@ public void ReorderRandomizer(int currentIndex, int nextIndex)
         }
 
         /// <summary>
-        /// Generates a random seed by hashing the current scenario iteration with a given base random seed
+        /// Generates and returns a new random seed
         /// </summary>
-        /// <param name="baseSeed">Used to offset the seed generator</param>
         /// <returns>The generated random seed</returns>
-        public uint GenerateRandomSeed(uint baseSeed = SamplerUtility.largePrime)
+        public uint NextRandomSeed()
```
StevenBorkman, line 341: Mentioned earlier, but this method name should probably reflect what is happening in here better

StevenBorkman (changes requested):
- Tutorial needs to be updated with your changes
- There should be a way for a user to choose to have a random seed generated for each run, this way each scenario will have different results. I suggest using a checkbox.
- Annotations should be added to perception json files that record the random seed (perhaps more info) of each simulation.

com.unity.perception/Runtime/Randomization/Samplers/SamplerTypes/NormalSampler.cs
sleal-unity: Replaced with NextRandomState() to better reflect that the next random value in a sequence of random states has been generated.

StevenBorkman (approved):
A new jira task should be created to allow a user to choose to use a random starting seed

PR #150, opened 2021-01-12
Added animation curve sampling + sampler deserialization bugfix + native sampling removal

mkamalza:
# Peer Review Information:
- Added AnimationCurveSampler, which returns random values according to a range and probability distribution denoted by a user provided AnimationCurve. 

- Fixed a sampler deserialization bug that could lead to type mismatch exceptions during deserialization.

- Minor fix to sampler serialization code to allow for serializing AnimationCurves.

## Dev Testing:
**Tests Added**: 
AnimationCurveSampler test added. Performs similarly to the Normal Sampler test.

## Checklist
- [+] - Updated docs
- [+] - Updated changelog


com.unity.perception/Runtime/Randomization/Samplers/ISampler.cs
```diff
@@ -28,5 +28,15 @@ public interface ISampler
...
+
+        /// <summary>
+        /// Used for performing sampler specific clean-up tasks (e.g. once the scenario is complete).
+        /// </summary>
+        void Cleanup();
+
+        /// <summary>
+        /// Used for performing sampler specific initialization tasks.
+        /// </summary>
+        void Initialize();
```
sleal-unity, line 40: While I'm all for optimization, I think enforcing an assumed lifecycle on samplers will get us into trouble. For example,  the split grammar uses samplers to generate random values in the editor to visualize intermediate results. The animation curve sampler would uninitialized in this scenario, since there is no game loop in the editor to call these lifecycle functions.

So I'd suggest the following:
- Check if the sampler is initialized in the sample function itself and cache any information there instead of calling the Initialize() method on all samplers from the scenario when the game loop begins
- The cleanup step as well can be replaced with a finalizer that disposes of the native array when the object is destroyed.
- Remove Initialize() and Cleanup() from the sampler interface

Example finalizer:
```
public class AnimationCurveSampler : ISampler
{
  public NativeArray<float> cachedData;

  ~AnimationCurveSampler()  // finalizer
  {
      cachedData.Dispose();
  }
}
```

sleal-unity (changes requested):

com.unity.perception/Runtime/Randomization/Samplers/SamplerTypes/AnimationCurveSampler.cs
```diff
@@ -0,0 +1,87 @@
...
+        /// </summary>
+        [Tooltip("Probability distribution curve used for this sampler. The X axis corresponds to the values this sampler will pick from, and the Y axis corresponds to the relative probability of the values. The relative probabilities (Y axis) do not need to max out at 1 as only the shape of the curve matters. The Y values cannot however be negative.")]
+        public AnimationCurve distributionCurve;
+
+        float[] m_IntegratedCurve;
+        bool m_CurveValid;
+        float m_StartTime;
+        float m_EndTime;
+        float m_Interval;
+        int m_NumOfSamples = 100;
```
sleal-unity, line 34: If m_NumOfSamples is a constant we should mark it as such. If it should be configurable by a user, I would make this setting public so they can initialize it in an initializer during construction of the sampler.

sleal-unity (changes requested):

com.unity.perception/Runtime/Randomization/Samplers/SamplerTypes/AnimationCurveSampler.cs
mkamalza: Good point. I made it public and added a tooltip to explain.

sleal-unity (approved):

PR #164, opened 2021-01-26
New capture options added to Perception Camera

mkamalza:
# Peer Review Information:

- Perception Camera now render intermediate frames between captures. 
  - The old "period" field is removed and instead the camera now requests a "simulationDeltaTime" from the simulation. 
  - Capture scheduling is now based on how many frames the camera "skips" between captures. All skipped frames are still rendered. 
  - Start time has been replaced by start frame.
  
- Perception Camera can now be triggered completely manually.
  - In this mode, the user can still choose to set the simulation's delta time by enabling a checkbox.  

- Multiple camera coordination still works similar to before. At each frame, the next frame's delta time is set based on the camera that will need to render or capture the earliest from now.

## Dev Testing:
**Tests Added**: 

Scheduling tests have been modified to reflect the new changes.

**Core Scenario Tested**: 

**At Risk Areas**: 

**Notes + Expectations**: 

## Checklist
- [ ] - Updated docs
- [ ] - Updated changelog


com.unity.perception/Runtime/GroundTruth/PerceptionCamera.cs
```diff
@@ -50,6 +43,61 @@ public partial class PerceptionCamera : MonoBehaviour
...
 
+        /// <summary>
+        /// Frame number at which this camera starts capturing.
+        /// </summary>
+        public int firstCaptureFrame = 0;
+
+        /// <summary>
+        /// Capture trigger modes for <see cref="PerceptionCamera"/>.
+        /// </summary>
+        public enum CaptureTriggerMode
```
JonathanHUnity, line 54: It is not good form to define enums inside classes if they are part of the public API. https://docs.microsoft.com/en-us/dotnet/standard/design-guidelines/nested-types

This looks like it should be moved out.

JonathanHUnity: Also, SimulationState should not depend on PerceptionCamera, since it should be able to be used by any sensor implementation.

com.unity.perception/Tests/Runtime/GroundTruthTests/DatasetCaptureSensorSchedulingTests.cs
```diff
@@ -61,40 +61,95 @@ public void TearDown()
...
+            {
                 var sensorData = m_TestHelper.GetSensorData(sensorHandle);
                 var sequenceTimeActual = m_TestHelper.CallSequenceTimeOfNextCapture(sensorData);
                 Assert.AreEqual(sequenceTimesExpected[i], sequenceTimeActual, 0.0001f);
+                yield return null;
             }
         }
 
+        [UnityTest]
+        public IEnumerator SequenceTimeOfNextCapture_WithInBetweenFrames_ReportsCorrectTime()
```
JonathanHUnity, line 92: Should add some tests around CaptureTriggerMode.Manual. Not sure if we would want tests with one scheduled and one manual as well as just one manual.

JonathanHUnity (changes requested):
Great overall, just a couple comments. One other idea for you - some users want this for physics and others want it for multi-frame rendering (like accumulation). For physics, rendering is not necessary and will add a lot of unnecessary overhead. How about we add an option to disable rendering during intermediate frames?

com.unity.perception/Runtime/GroundTruth/PerceptionCamera.cs
mkamalza: I moved the enum to DatasetCapture.cs, and out of any other class. Seems like a sensible place for it as sensor stuff is defined there.

com.unity.perception/Tests/Runtime/GroundTruthTests/DatasetCaptureSensorSchedulingTests.cs
mkamalza: I added a couple of scheduling tests for triggering the manual capture and checking if the capture time has been correctly set.

mkamalza: I also modified the dual sensor test and added one manual sensor into the mix. This tests for delta times and capture flags being correctly set when coordinating multiple sensors and some of the are manual.

JonathanHUnity (approved):

PR #165, opened 2021-01-27
Output dir shortcut

mkamalza:
# Peer Review Information:

* Added two buttons to the Perception Camera's inspector for opening the latest output folder and copying its path to clipboard.
  * Although the buttons are added to the Perception Camera, the directory is the same across all sensors for each run. It changes on each run. But since most (if not all) of our users seem to use a single camera, this solution should be OK for now.
  
## Dev Testing:
**Tests Added**: 

**Core Scenario Tested**: 

**At Risk Areas**: 

**Notes + Expectations**: 

## Checklist
- [ ] - Updated docs
- [ ] - Updated changelog


JonathanHUnity (approved):
Sweet

PR #166, opened 2021-01-27
RandomzierTagManager.Query<T>() now returns tags directly instead of GameObjects

sleal-unity:
# Peer Review Information:
This PR enables the RandomzierTagManager.Query<T>() method to return typed RandomizerTags directly instead of the GameObjects said tags are attached to.

## Checklist
- [X] - Updated docs
- [X] - Updated changelog


StevenBorkman (approved):
Looks good! Actually cleaned the code up some too.

PR #177, opened 2021-01-31
Removed randomization tooling from the experimental namespace

sleal-unity:
# Peer Review Information:
This PR removes the randomization tooling from the experimental namespace.

Note: I noticed that we include some references to assemblies that are no longer present in the perception package (PathCreator and QuickGraph) so I used this PR as an opportunity to remove them. I also sorted the order of the existing asmdef references while playing with Rider's "Sort Lines" code editing feature.

## Checklist
- [X] - Updated docs
- [X] - Updated changelog

### DIFF
--- a/com.unity.perception/Runtime/Randomization/Parameters/ParameterTypes/CategorialParameters/GameObjectParameter.cs
+++ b/com.unity.perception/Runtime/Randomization/Parameters/ParameterTypes/CategorialParameters/GameObjectParameter.cs
@@ -1,4 +1,5 @@
 ﻿using System;
+using UnityEngine.Scripting.APIUpdating;
 
 namespace UnityEngine.Perception.Randomization.Parameters
 {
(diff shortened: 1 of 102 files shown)


com.unity.perception/Runtime/Randomization/Parameters/ParameterTypes/CategorialParameters/GameObjectParameter.cs
```diff
@@ -1,6 +1,6 @@
 ﻿using System;
 
-namespace UnityEngine.Experimental.Perception.Randomization.Parameters
+namespace UnityEngine.Perception.Randomization.Parameters
```
JonathanHUnity, line 3: Since there are many users of these APIs now, we should add [MovedFrom] decorations to the public types so that the script updater will automatically update users' code when they update to the latest package. For more, see https://confluence.unity3d.com/display/DEV/Move+type+to+another+namespace

sleal-unity: Cool! I've added this attribute to all our public API's.

JonathanHUnity (approved):

PR #184, opened 2021-02-03
Pose config rename

StevenBorkman:
# Peer Review Information:
Renamed pose label config classes & files to be consistent. Fixed assorted stability issues with grabbing correct pose label.

## Checklist
- [x] - Updated docs
- [x] - Updated changelog


com.unity.perception/Runtime/GroundTruth/Labelers/KeyPointLabeler.cs
```diff
@@ -72,7 +73,8 @@ public KeyPointLabeler(IdLabelConfig config, KeyPointTemplate template)
         /// <summary>
         /// Array of animation pose labels which map animation clip times to ground truth pose labels.
         /// </summary>
-        public AnimationPoseLabel[] poseStateConfigs;
+        [FormerlySerializedAs("poseStateConfigs")]
```
sleal-unity, line 76: I'm not sure if we'll need this attribute since the package hasn't released yet

sleal-unity (changes requested):

com.unity.perception/Runtime/GroundTruth/Labelers/KeyPointLabeler.cs
StevenBorkman: Good catch. I meant to remove that but put it there temporarily so my test files weren't broken

sleal-unity (approved):

PR #190, opened 2021-02-06
Update KeyPointLabeler.cs

mkamalza:
Tiny fix on behalf of Steve B

# Peer Review Information:

This PR changes the reported state for manually added joints from 1 to 2, which is the correct state for them. (change made on behalf of Steve B)

## Dev Testing:
**Tests Added**: 

**Core Scenario Tested**: 

**At Risk Areas**: 

**Notes + Expectations**: 

## Checklist
- [ ] - Updated docs
- [ ] - Updated changelog

### DIFF
--- a/com.unity.perception/Runtime/GroundTruth/Labelers/KeyPointLabeler.cs
+++ b/com.unity.perception/Runtime/GroundTruth/Labelers/KeyPointLabeler.cs
@@ -165,7 +165,10 @@ namespace UnityEngine.Perception.GroundTruth
             /// </summary>
             public float y;
             /// <summary>
-            /// The state of the point, 0 = not present, 1 = keypoint is present
+            /// The state of the point,
+            /// 0 = not present,
+            /// 1 = keypoint is present but not visible,
+            /// 2 = keypoint is present and visible
             /// </summary>
             public int state;
         }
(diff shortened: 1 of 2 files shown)


StevenBorkman (approved):

com.unity.perception/Runtime/GroundTruth/Labelers/KeyPointLabeler.cs
```diff
@@ -309,7 +309,7 @@ void ProcessEntity(Labeling labeledEntity)
                     keyPoints[idx].index = idx;
                     keyPoints[idx].x = loc.x;
                     keyPoints[idx].y = loc.y;
-                    keyPoints[idx].state = 1;
+                    keyPoints[idx].state = 2;
```
sleal-unity, line 312: Why would this be 2 instead of 1 or 0? Based on the comments in the Keypoint class definition, this state value can only be either 1 or 0 based on whether the Keypoint is present or not. Actually, I'm not sure why this isn't a boolean. There's another line right above this line that also sets state to 2, and I'm not sure why that it.

sleal-unity (changes requested):

com.unity.perception/Runtime/GroundTruth/Labelers/KeyPointLabeler.cs
mkamalza: So according to Steve, 1 means the joint is present but not visible and 2 means present and visible. These are based on COCO. 0 means not present. 

Steve said 1 is not actually implemented yet (future work). When you add a joint manually, it should always be 2 since it exists.

sleal-unity (approved):

PR #209, opened 2021-02-17
Fixed parameter element editor

StevenBorkman:
Now allows 'Add Folder' to append on already present elements

# Peer Review Information:
See above

## Dev Testing:
**Tests Added**: 

**Core Scenario Tested**: 

**At Risk Areas**: 

**Notes + Expectations**: 

## Checklist
- [x] - Updated docs
- [x] - Updated changelog

### DIFF
--- a/com.unity.perception/CHANGELOG.md
+++ b/com.unity.perception/CHANGELOG.md
@@ -23,6 +23,8 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 
 Fixed ground truth not properly produced when there are other disabled PerceptionCameras present. Note: this does not yet add support for multiple enabled PerceptionCameras.
 
+Fixed an issue preventing a user from adding more parameters to a parameter list with the 'Add Folder' button. 'Add Folder' now appends the contents of the new folder on the list instead of replacing its contents.
+
 ## [0.7.0-preview.2] - 2021-02-08
 
 ### Upgrade Notes
--- a/com.unity.perception/CHANGELOG.md
+++ b/com.unity.perception/CHANGELOG.md
@@ -23,7 +23,9 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 
 Fixed ground truth not properly produced when there are other disabled PerceptionCameras present. Note: this does not yet add support for multiple enabled PerceptionCameras.
 
-Fixed an issue preventing a user from adding more parameters to a parameter list with the 'Add Folder' button. 'Add Folder' now appends the contents of the new folder on the list instead of replacing its contents.
+Fixed an issue preventing a user from adding more options to a Categorical Parameter's list of options with the 'Add Folder' button. 'Add Folder' now correctly appends the contents of the new folder on the list.
+
+Fixed a bug where uniform probabilities were not properly reset upon adding or removing options from a Categorical Parameter's list of options.
 
 ## [0.7.0-preview.2] - 2021-02-08
 
--- a/com.unity.perception/Editor/Randomization/VisualElements/Parameter/ParameterElement.cs
+++ b/com.unity.perception/Editor/Randomization/VisualElements/Parameter/ParameterElement.cs
@@ -87,13 +87,28 @@ namespace UnityEditor.Perception.Randomization
                     // First delete sets option to null, second delete removes option
                     var numOptions = optionsProperty.arraySize;
                     optionsProperty.DeleteArrayElementAtIndex(i);
+
                     if (numOptions == optionsProperty.arraySize)
+                    {
                         optionsProperty.DeleteArrayElementAtIndex(i);
+                    }
+
+                    ResetProbabilities();
 
                     m_SerializedProperty.serializedObject.ApplyModifiedProperties();
+
                     listView.itemsSource = categoricalParameter.probabilities;
                     listView.Refresh();
                 };
+            }
+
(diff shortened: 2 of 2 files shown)


com.unity.perception/Editor/Randomization/VisualElements/Parameter/ParameterElement.cs
```diff
@@ -87,15 +87,30 @@ void BindItem(VisualElement e, int i)
                     // First delete sets option to null, second delete removes option
                     var numOptions = optionsProperty.arraySize;
                     optionsProperty.DeleteArrayElementAtIndex(i);
+
                     if (numOptions == optionsProperty.arraySize)
+                    {
                         optionsProperty.DeleteArrayElementAtIndex(i);
+                    }
+
+                    ResetProbabilities();
```
mkamalza, line 96: I think the reset line should only be called if `uniform` is enabled, right? Otherwise we keep the probabilities as they are.

com.unity.perception/Editor/Randomization/VisualElements/Parameter/ParameterElement.cs
```diff
@@ -114,6 +129,8 @@ void BindItem(VisualElement e, int i)
                         break;
                 }
 
+                ResetProbabilities();
```
mkamalza, line 132: Same comment for this one.

com.unity.perception/Editor/Randomization/VisualElements/Parameter/ParameterElement.cs
```diff
@@ -129,17 +146,20 @@ void BindItem(VisualElement e, int i)
...
                     for (var i = 0; i < categories.Count; i++)
                     {
-                        var optionProperty = optionsProperty.GetArrayElementAtIndex(i);
-                        var probabilityProperty = probabilitiesProperty.GetArrayElementAtIndex(i);
+                        var optionProperty = optionsProperty.GetArrayElementAtIndex(optionsIndex + i);
                         optionProperty.objectReferenceValue = categories[i];
-                        probabilityProperty.floatValue = uniformProbability;
                     }
 
+                    ResetProbabilities();
```
mkamalza, line 161: And this one

com.unity.perception/CHANGELOG.md
```diff
@@ -23,6 +23,8 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 
 Fixed ground truth not properly produced when there are other disabled PerceptionCameras present. Note: this does not yet add support for multiple enabled PerceptionCameras.
 
+Fixed an issue preventing a user from adding more parameters to a parameter list with the 'Add Folder' button. 'Add Folder' now appends the contents of the new folder on the list instead of replacing its contents.
```
mkamalza, line 26:
```suggestion
Fixed an issue preventing a user from adding more options to a Categorical Parameter's list of options with the 'Add Folder' button. 'Add Folder' now correctly appends the contents of the new folder on the list.

Fixed a bug where uniform probabilities were not properly reset upon adding or removing options from a Categorical Parameter's list of options.
```

mkamalza (changes requested):
couple of minor things

com.unity.perception/Editor/Randomization/VisualElements/Parameter/ParameterElement.cs
StevenBorkman: I'm not sure on that, unless we default all new entries to 0% probability. If someone has their probabilities custom set and then they add in two new parameters then, either the new ones have to be set to 0 out of the box, or we need to rebalance. I'm not sure what the right answer is here, or if there is one. Thoughts?

mkamalza: I see your point. My thoughts are, if the user has manually set probabilities, they are aware of the feature and how to use it. So when they add new options they would probably know they need to set probabilities for them. 
On the other hand, resetting the probabilities can potentially remove numbers that the user has carefully set.

mkamalza: The default behaviour when uniform is disabled and a single option is added is also to set the probability to 0 right?

StevenBorkman: I'll have to get a clean branch to verify but I believe it recalculates

mkamalza: So, the current behaviour is a bit inconsistent depending on whether there are previous options already added, and whether a single option or a folder is added next. I think what we could when uniform is **disabled**:
- When a folder is added, all the new options get a probability equal to 1/(number of new options). This is how it works now. It does lead to the sum of all options being more than 1, but that's ok, it is normalized later.
- When a single option is added, we set its probability to 0. Ideally, whenever an option has a 0 probability we could also show a warning box saying there are items with 0 probability. The current behaviour is to set this to the probability of the last element previously in the list, or 0 if the list is empty.

I am mainly concerned about changing probabilities that are already set by the user. That would not be a good UX.

StevenBorkman: Ok, reworked it so that added entries default to probability of 0 if uniform is unchecked

mkamalza (approved):
thanks!

PR #211, opened 2021-02-18
Fixing error with non-public serialized fields on randomizers

JonathanHUnity:
# Peer Review Information:
Fixing bug in UIElementsEditorUtilities on types that have non-public serialized fields. This issue was hit by a user.

This also makes UIElementsEditorUtilities an internal type, as I would suggest it does not contribute to the core functionality of the package. Feel free to discuss in the comments, as it may be important to expose this in some way to users.

Finally, this adds some error checking for indications that a randomizer script is missing and presents a helpful message to the user instead of just throwing exceptions.

## Dev Testing:
**Tests Added**:  Added unit tests for UIElementsEditorUtilities.

**Core Scenario Tested**: 

**At Risk Areas**: 

**Notes + Expectations**: 

## Checklist
- [ ] - Updated docs
- [x] - Updated changelog


JonathanHUnity:
image

com.unity.perception/Editor/Randomization/Utilities/UIElementsEditorUtilities.cs
```diff
@@ -66,7 +67,7 @@ public static PropertyField CreatePropertyField(SerializedProperty iterator, Typ
         {
             var propertyField = new PropertyField(iterator.Copy());
             propertyField.Bind(iterator.serializedObject);
-            var originalField = parentPropertyType.GetField(iterator.name);
+            var originalField = parentPropertyType.GetField(iterator.name, BindingFlags.Public | BindingFlags.NonPublic | BindingFlags.Instance );
             var tooltipAttribute = originalField.GetCustomAttributes(true)
```
mkamalza, line 71: unrelated to new code, but `originalField` could be null here. Maybe add a null check?

com.unity.perception/Tests/Editor/UIElementsEditorUtilitiesTests.cs
```diff
@@ -0,0 +1,47 @@
...
+        internal string internalField;
+    }
+    [Test]
+    [TestCase(typeof(TestType), nameof(TestType.publicField))]
+    [TestCase(typeof(TestType), nameof(TestType.publicFieldWithTooltip), "A tooltip")]
+    [TestCase(typeof(TestType), "privateField", "A tooltip")]
+    [TestCase(typeof(TestType), "protectedField", "A tooltip")]
+    [TestCase(typeof(TestType), "protectedInternalField", "A tooltip")]
+    [TestCase(typeof(TestType), "internalField", "A tooltip")]
+    public void CreatePropertyField_ReturnsCorrectPropertyField_ForTypeAndField(Type type, string field,
```
mkamalza, line 36: I love this test!

mkamalza (approved):
Looks good! just a couple of tiny things.

Regarding UIElementsEditorUtilities being public or not, I don't see why it should be public, but would like @sleal-unity's opinion too as he originally created this. 

P.S. I missed capitalization for "Randomizer" in the red UI message in my review suggestions.

sleal-unity (commented):
Yeah, you can make UIElementsEditorUtilities private. I had some justification in mind when I made it public that is actually very weak on second thought. So yeah, it should be private.

sleal-unity (approved):

com.unity.perception/Editor/Randomization/Utilities/UIElementsEditorUtilities.cs
JonathanHUnity: Could it? If it is ever null that would be a bug with this code, unless I'm missing something

mkamalza: It can happen if there are stale serialized properties in the assets that do not exist in the classes any more, right? Might be a good idea to just ignore them if we come across them.

JonathanHUnity: Those fields should be removed from the SerializedObject at import time. Just to be sure, I tested it out by adding a field and then removing it, and I didn't see any issues. Good thing to try though 👍🏻

mkamalza: Aa yes, you're right. Neat!

PR #213, opened 2021-02-19
Scenario lifecycle hooks update

sleal-unity:
# Peer Review Information:

**NOTICE**: I found two commits I made 4 days ago using a different GitHub user name that I've amended in this PR. Everything should merge properly, but you'll see an interestingly long commit history being listed here, when in reality, there's just one commit actually being made.

This PR refactors the existing scenario lifecycle hooks to make custom scenarios easier to derive and prepare the scenario class to enable more ADR-like data streaming workflows. This PR also adds the PerceptionScenario abstract class to better distinguish the scenario's coordinator responsibilities from basic perception functionality.

## Dev Testing:
**Tests Added**: 
Existing tests have been updated to use the new lifecycle hooks

## Checklist
- [X] - Updated docs
- [X] - Updated changelog


com.unity.perception/Runtime/Randomization/Scenarios/PerceptionScenario.cs
```diff
@@ -0,0 +1,94 @@
...
+        protected override void OnComplete()
+        {
+            DatasetCapture.ResetSimulation();
+            Manager.Instance.Shutdown();
+        }
+
+        /// <inheritdoc/>
+        protected override void OnIdle()
+        {
+            if (Manager.FinalUploadsDone)
```
JonathanHUnity, line 81: This should not be necessary. Calling Quit() will automatically call Manager.Instance.Shutdown() and should terminate synchronously.

com.unity.perception/Runtime/Randomization/Scenarios/PerceptionScenario.cs
```diff
@@ -0,0 +1,94 @@
...
+                Quit();
+        }
+
+        /// <summary>
+        /// Used to report a scenario iteration as a perception metric
+        /// </summary>
+        protected struct IterationMetricData
```
JonathanHUnity, line 88: why protected? Nested types in the public API should be avoided.

com.unity.perception/Runtime/Randomization/Scenarios/PerceptionScenario.cs
sleal-unity: Ahh, I'll need to test this again then. In the past, Devin and I ran into a problem with how the final uploads took a few frames to complete before we could quit, or else the simulation would not save all the recorded data properly.

com.unity.perception/Runtime/Randomization/Scenarios/PerceptionScenario.cs
sleal-unity: My thought was that someone deriving this class would maybe want to either not record this metric or potentially rearrange the timing a bit, and therefore require the class to be accessible in some way so they could use it in their derived class.

sleal-unity: What do you think about just moving it outside of the PerceptionScenarioClass and making it public?

com.unity.perception/Runtime/Randomization/Scenarios/PerceptionScenario.cs
sleal-unity: Yep, this totally works. The scenario now quits within the OnComplete() hook

com.unity.perception/Runtime/Randomization/Scenarios/PerceptionScenario.cs
sleal-unity: I just made it private. It's not a big deal.

PR #220, opened 2021-02-23
Fix pose positions

JonathanHUnity:
# Peer Review Information:
This change fixes a bug where keypoint positions would be wrong on new objects and objects off screen when animations are playing.

This was happening because for these objects, transforms are only updated by the animation system during camera culling, and the positions were being recorded before culling occurred. This change adds a new event `CameraLabeler.OnEndRendering` to allow the keypoint labeler to instead record positions after rendering.

## Dev Testing:
**Tests Added**: Added tests for the new event and for animations + keypoint labeling.

**Core Scenario Tested**:  Tested with test scene and with our human pose project.

**At Risk Areas**: 

**Notes + Expectations**: 

## Checklist
- [ ] - Updated docs
- [ ] - Updated changelog


com.unity.perception/Runtime/GroundTruth/PerceptionCamera.cs
```diff
@@ -424,21 +426,26 @@ void OnSimulationEnding()
...
+        {
+            if (!ShouldCallLabelers(cam, m_LastFrameEndRendering))
+                return;
 
+            m_LastFrameEndRendering = Time.frameCount;
+            CallOnLabelers(l => l.InternalOnEndRendering());
+        }
+
+
+        private void CallOnLabelers(Action<CameraLabeler> action)
```
StevenBorkman, line 447: I'm not sure about this name, how about something like: NotifyLabelers(Action<CameraLabeler> action)

StevenBorkman (changes requested):
Another comment: we should probably add these assets to the URP project also

Beyond that, good find on that bug!

StevenBorkman (approved):

PR #227, opened 2021-03-01
Cloud rendering support

PriyeshWani:
# Peer Review Information:
Updated the RunInUnitySimulationWindow script to add support for cloudrendering build.
The changes will build for CloudRendering build target when the user is on that build target and check for sysparam selection.
It will prompt if the right sysparam is not selected by the developer (with GPU) from the dropdown.

## Dev Testing:
Verified locally by running simulation with cloudrendering build target.
Verfied the dialog with different sysparam selection

**Tests Added**: 

**Core Scenario Tested**: 

**At Risk Areas**: 
Unity Simulation Build and Run

**Notes + Expectations**: 
- When switched to cloud rendering build target, Build and Run option should create a headless build with cloud rendering build target.
- If appropriate sysparam is not selected, the editor should prompt a dialog for appropriate sysparam selection.

## Checklist
- [ ] - Updated docs
- [ ] - Updated changelog

### DIFF
--- a/com.unity.perception/Editor/Randomization/Editors/RunInUnitySimulationWindow.cs
+++ b/com.unity.perception/Editor/Randomization/Editors/RunInUnitySimulationWindow.cs
@@ -35,6 +35,8 @@ namespace UnityEditor.Perception.Randomization
         Label m_PrevExecutionIdLabel;
         RunParameters m_RunParameters;
 
+        private string supportedGPU = "NVIDIA";
+
         [MenuItem("Window/Run in Unity Simulation")]
         static void ShowWindow()
         {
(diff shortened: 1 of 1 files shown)


com.unity.perception/Editor/Randomization/Editors/RunInUnitySimulationWindow.cs
```diff
@@ -166,6 +168,15 @@ static string IncrementRunName(string runName)
 
         async void RunInUnitySimulation()
         {
+            #if PLATFORM_CLOUD_RENDERING
```
JonathanHUnity, line 171: What is the intended use case here? Only internal devs will have access to CloudRendering build target, so this enhancement doesn't appear to be for users. Is that your intent with this change? I was under the impression you would also add support for selecting an existing build id so that we can use builds uploaded from Cloud Build.

JonathanHUnity (changes requested):

com.unity.perception/Editor/Randomization/Editors/RunInUnitySimulationWindow.cs
PriyeshWani: That is correct, for now, only internal devs and on-prem customers who have access to the cloud rendering build target will have an option to switch to cloud rendering and then do the build and run on usim. The cloudbuild workflow is still not finalized. There are a lot of friction points with that workflow and would require some work on the cloud build side as well. For now the focus is on on-prem customers for the advantaged SKU and internal usage.

JonathanHUnity: Ok. Can follow up offline

JonathanHUnity (approved):

PR #235, opened 2021-03-04
Keypoint tolerance

StevenBorkman:
# Peer Review Information:
Added some tolerance to keypoint check and better code comments on surrounding code

## Dev Testing:
**Tests Added**: 

**Core Scenario Tested**: 

**At Risk Areas**: 

**Notes + Expectations**: 

## Checklist
- [x] - Updated docs
- [x] - Updated changelog


com.unity.perception/Runtime/GroundTruth/Labelers/KeypointLabeler.cs
```diff
@@ -108,14 +107,62 @@ bool AreEqual(Color32 lhs, Color32 rhs)
...
+        }
+
+        bool PixelsMatch(int x, int y, Color32 idColor, (int x, int y) dimensions, NativeArray<Color32> data)
+        {
+            var h = dimensions.y - y;
+            var pixelColor = data[h * dimensions.x + x];
+            return AreEqual(pixelColor, idColor);
+        }
+
+        static int s_PixelTolerance = 2;
```
JonathanHUnity, line 122: Can you explain why we need '2' here? a 5x5 kernel is a lot of ops, and it feels like we might be covering up a bug in some other math here.

com.unity.perception/Runtime/GroundTruth/Labelers/KeypointLabeler.cs
```diff
@@ -108,14 +107,62 @@ bool AreEqual(Color32 lhs, Color32 rhs)
...
+
+            var pixelOnScreen = false;
+
+            for (var x = centerX - s_PixelTolerance; x <= centerX + s_PixelTolerance; x++)
+            {
+                for (var y = centerY - s_PixelTolerance; y <= centerY + s_PixelTolerance; y++)
+                {
+                    if (!PixelOnScreen(x, y, dimensions)) continue;
+
+                    pixelOnScreen = true;
```
JonathanHUnity, line 147: This is an awkward thing to document - a keypoint is on the screen if it is outside the image by 2 pixels?

com.unity.perception/Runtime/GroundTruth/Labelers/KeypointLabeler.cs
```diff
@@ -108,14 +107,62 @@ bool AreEqual(Color32 lhs, Color32 rhs)
...
+        int DetermineKeypointState(Keypoint keypoint, Color32 instanceIdColor, (int x, int y) dimensions, NativeArray<Color32> data)
+        {
+            if (keypoint.state == 0) return 0;
+
+            var centerX = (int)keypoint.x;
+            var centerY = (int)keypoint.y;
+
+            var pixelOnScreen = false;
+
+            for (var x = centerX - s_PixelTolerance; x <= centerX + s_PixelTolerance; x++)
```
JonathanHUnity, line 141: Two suggestions - check the center pixel first to early out, as in most cases we would expect this to be the first one. Second, iterate over x on the inner loop to improve cache coherance.

JonathanHUnity (changes requested):
Couple of things to look into here.

com.unity.perception/Runtime/GroundTruth/Labelers/KeypointLabeler.cs
StevenBorkman: yep, needed to round instead of just casting.

com.unity.perception/Runtime/GroundTruth/Labelers/KeypointLabeler.cs
StevenBorkman: Implemented the second suggestion of flipping x and y loops. As for the other suggestion, I think the 95% use case, we are going to either hit every pixel in the block or miss everyone, it is only real outside use cases, like the exact corner of the box in our test scene where this will be issue. So I think the single case in the code will keep the code more clean and readable instead of adding a special case.

JonathanHUnity (approved):

PR #236, opened 2021-03-05
Quick fix for a out of range error that could occur

StevenBorkman:
A quick fix for an index out of bounds error that could occur if a keypoint template file was not set up properly

# Peer Review Information:

## Dev Testing:
**Tests Added**: 

**Core Scenario Tested**: 

**At Risk Areas**: 

**Notes + Expectations**: 

## Checklist
- [x] - Updated docs
- [x] - Updated changelog

### DIFF
--- a/com.unity.perception/CHANGELOG.md
+++ b/com.unity.perception/CHANGELOG.md
@@ -65,6 +65,8 @@ Fixed a bug where uniform probabilities were not properly reset upon adding or r
 
 Fixed keypoints being reporeted in wrong locations on the first frame an object is visible.
 
+Fixed an out of range error if a keypoint template skeleton relies on a joint that is not available.
+
 ## [0.7.0-preview.2] - 2021-02-08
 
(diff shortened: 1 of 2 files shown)


JonathanHUnity (approved):
Should be awhile to write a test for this. Also, is there any impact on the json dataset with this setup?

StevenBorkman:
> Should be awhile to write a test for this. Also, is there any impact on the json dataset with this setup?

This will not impact json file format

JonathanHUnity:
Sorry, I meant it would be good to add a test. What is the dataset output with one of these non matching joint labels?

StevenBorkman:
> Sorry, I meant it would be good to add a test. What is the dataset output with one of these non matching joint labels?

So this is really only for visualization, the fix is in our visualization code. Our keypoint labeler does not care or really even know about the skeletal connections. A user who ingests our template file would have to handle this themselves, unless we were to write some sort of template validation tool

PR #240, opened 2021-03-08
Remove entities dependency

sleal-unity:
# Peer Review Information:
This PR removes the entities package dependency to avoid an API updater bug that triggers when opening the project settings window on 2020.2

Note that the API updater bug still occurs on URP 2020.2.6f1 even with this PR, but the source of the issue is no longer the entities package. Upon opening the project settings window, the following line is repeatedly updated within the Runtime/SRPSupport.cs file of the com.unity.simulation.capture package:
```
-        rtid = additionalCameraData.scriptableRenderer.cameraDepthTarget;
+        rtid = additionalCameraData.scriptableRenderer.cameraDepth;^M
```

On a separate note, some key changes to review in this PR include:
- The removal of the entities package dependency
- The new LabelManager and PerceptionUpdater classes (replaces GroundTruthLabelSetupSystem and SimulationManagementComponentSystem)
- The new GroundTruth/RenderPasses directory that organizes all our custom render pass scripts
- The URP scriptable render passes that have been moved from the GroundTruthRenderer script into their own scripts
- The inclusion of the com.unity.burst and com.unity.collections dependencies

Author note: I took this as an opportunity to clean up some code in the GroundTruth section of the package. You will see a lot of minor changes like field renamings, rearranged code, and type definitions replaced with "var".

## Dev Testing:
**Tests Updated**: 
Removing the entities dependency required adding a specific dependency version for com.unity.burst and com.unity.collections that was compatible with 2019.4. A few tests have been updated to accommodate the lack of "enumerable" support for NativeArrays in earlier versions of the collections package.

## Checklist
- [X] - Updated changelog


com.unity.perception/Runtime/GroundTruth/Labelers/SemanticSegmentationLabeler.cs
```diff
@@ -32,13 +29,13 @@ public override string description
...
                    "Semantic segmentation images are saved to the dataset in PNG format. " +
-                   "Please note that only one " + this.GetType().Name + " can render at once across all cameras.";
+                   "Please note that only one " + GetType().Name + " can render at once across all cameras.";
             protected set {}
         }
 
         const string k_SemanticSegmentationDirectory = "SemanticSegmentation";
         const string k_SegmentationFilePrefix = "segmentation_";
-        internal string m_SemanticSegmentationDirectory;
+        internal string semanticSegmentationDirectory;
```
JonathanHUnity, line 38: Why remove the prefix?

com.unity.perception/Runtime/GroundTruth/Labeling/LabelManager.cs
```diff
@@ -0,0 +1,167 @@
...
+        public IEnumerable<Labeling> registeredLabels => m_RegisteredLabels;
+
+        /// <summary>
+        /// Registers all pending labels.
+        /// Called once per frame during LateUpdate by the <see cref="PerceptionUpdater"/>.
+        /// </summary>
+        public void RegisterPendingLabels()
+        {
+            if (m_RegisteredLabels.Count == 0)
+                m_CurrentObjectIndex = k_StartingIndex;
```
JonathanHUnity, line 34: 'NextObjectIndex' is the standard way of naming a value like this. Why 'current'?

com.unity.perception/Runtime/GroundTruth/Labeling/LabelManager.cs
```diff
@@ -0,0 +1,167 @@
...
+        public bool Deactivate(IGroundTruthGenerator generator)
+        {
+            return m_ActiveGenerators.Remove(generator);
+        }
+
+        /// <summary>
+        /// Registers a labeling component
+        /// </summary>
+        /// <param name="labeledObject">the component to register</param>
+        public void Register(Labeling labeledObject)
```
JonathanHUnity, line 87: Why should this be public?

com.unity.perception/Runtime/GroundTruth/Labeling/LabelManager.cs
```diff
@@ -0,0 +1,167 @@
...
+            m_LabelsPendingRegistration.Remove(labeledObject);
+            m_RegisteredLabels.Remove(labeledObject);
+        }
+
+        /// <summary>
+        /// Refresh ground truth generation for the labeling of a particular GameObject. This is necessary when the
+        /// list of labels changes or when renderers or materials change on objects in the hierarchy.
+        /// </summary>
+        /// <param name="labeledObject">the component to refresh</param>
+        public void RefreshLabeling(Labeling labeledObject)
```
JonathanHUnity, line 107: Unnecessary for this to be public, since the API already exists on Labeling itself.

com.unity.perception/Runtime/GroundTruth/Labeling/LabelManager.cs
```diff
@@ -0,0 +1,167 @@
...
+        public void Register(Labeling labeledObject)
+        {
+            m_LabelsPendingRegistration.Add(labeledObject);
+        }
+
+        /// <summary>
+        /// Unregisters a labeling component
+        /// </summary>
+        /// <param name="labeledObject">the component to unregister</param>
+        public void Unregister(Labeling labeledObject)
```
JonathanHUnity, line 96: Why public?

com.unity.perception/Runtime/GroundTruth/PerceptionCamera.cs
```diff
@@ -157,155 +162,82 @@ public bool RemovePersistentSensorData(string key)
...
-            // issue for this (aisv-779) and have notified the engine team about this.
-            anyVisualizing = true;
-
+            // if we are utilizing async readback and we have to flip our captured image.
+            // We have created a jira issue for this (aisv-779) and have notified the engine team about this.
             if (m_ShowingVisualizations)
-                CaptureOptions.useAsyncReadbackIfSupported = !anyVisualizing;
-        }
-
-        private Vector2 scrollPosition;
```
JonathanHUnity, line 267: These large code movements make an already complex PR significantly harder to review. I realize I am also guilty of this on a fairly regular basis, but do be cautious about where you make these kinds of changes. Preferably they would be saved for a separate PR

com.unity.perception/Runtime/GroundTruth/Labeling/LabelManager.cs
```diff
@@ -0,0 +1,167 @@
+﻿using System.Collections.Generic;
+using UnityEngine.Perception.Randomization.Randomizers;
+
+namespace UnityEngine.Perception.GroundTruth
+{
+    /// <summary>
+    /// Manages the registration of <see cref="Labeling"/> components
+    /// </summary>
+    public class LabelManager
```
JonathanHUnity, line 9: I've kicked off a performance test run on this branch. We should compare the results to the lastest on master to look for significant performance changes.

JonathanHUnity: https://yamato.cds.internal.unity3d.com/jobs/907-com.unity.perception/tree/remove-entities-dependency/.yamato%252Fupm-ci-full.yml%2523all_tests_nightly_perf/5697974/job

com.unity.perception/Tests/Runtime/GroundTruthTests/SegmentationGroundTruthTests.cs
```diff
@@ -83,7 +83,8 @@ public enum SegmentationKind
                 if (frameStart == null || frameStart > frameCount) return;
 
                 timesSegmentationImageReceived++;
-                CollectionAssert.AreEqual(Enumerable.Repeat(expectedPixelValue, data.Length), data);
+                for (var i = 0; i < data.Length; i++)
```
JonathanHUnity, line 86: This check is not exactly the same as CollectionAssert.AreEqual, as it does not ensure the lengths of the collections are the same. It is also less explicit. I would suggest just calling .ToArray() on the NativeArray instead.

com.unity.perception/Runtime/GroundTruth/Labeling/LabelManager.cs
```diff
@@ -0,0 +1,167 @@
...
+    {
+        /// <summary>
+        /// Returns the active LabeledObjectsManager instance
+        /// </summary>
+        public static LabelManager singleton { get; } = new LabelManager();
+
+        const uint k_StartingIndex = 1;
+        uint m_CurrentObjectIndex = k_StartingIndex;
+        List<IGroundTruthGenerator> m_ActiveGenerators = new List<IGroundTruthGenerator>();
+        LinkedHashSet<Labeling> m_LabelsPendingRegistration = new LinkedHashSet<Labeling>();
```
JonathanHUnity, line 19: LinkedHashSet is clever, but it allocates GC memory on every insertion due to the linked list. We should be able to do better than this. I think this is worth fixing before landing the PR because otherwise it adds to our growing GC tech debt.

JonathanHUnity (changes requested):
Overall very good work. Left a handful of pieces of feedback to address, but no major concerns.

com.unity.perception/Runtime/GroundTruth/Labeling/LabelManager.cs
sleal-unity: This was copy pasted code from the old system, so I didn't think to hard about changing the naming conventions. I agree and will change this. Good catch.

com.unity.perception/Runtime/GroundTruth/Labeling/LabelManager.cs
sleal-unity: I'm open to suggestions, but LinkedHashSet is great in that it provides O(1) lookup, insertion, and removal. The LabelManager will require the capability to remove arbitrary registered LabelingComponents from it's registry, so the only alternative data structure I can think of is a normal array in which removal would be an O(n^2) operation.

We've haven't seen any performance issues when employing this data structure with the TagManager either. The use of the AddComponentObject API in the previous Labeling and GroundTruthLabelSetupSystem implementation nullified any real-benefit from using ECS. Here is a quote from the [Unity ECS documentation](https://docs.unity3d.com/Packages/com.unity.entities@0.0/api/Unity.Entities.EntityManager.html) to clarify:
> Accessing data in a managed object forfeits many opportunities for increased performance. Adding managed objects to an entity should be avoided or used sparingly. Adding a component changes an entity's archetype and results in the entity being moved to a different chunk.

I think implementing a more complex alternative Label registration data structure would be preemptively optimizing a solution that has no data supporting that it is under-performing. Actually, I would be inclined to believe that the new LinkedHashSet implementation is more performant than the previous ECS implementation.

com.unity.perception/Tests/Runtime/GroundTruthTests/SegmentationGroundTruthTests.cs
sleal-unity: Eyyy, this is why they pay ya the big bucks Jon. I'll implement your suggestion.

com.unity.perception/Runtime/GroundTruth/Labeling/LabelManager.cs
sleal-unity: Good catch, I'll change this to internal.

com.unity.perception/Runtime/GroundTruth/Labeling/LabelManager.cs
sleal-unity: Yep, this should be internal.

com.unity.perception/Runtime/GroundTruth/Labeling/LabelManager.cs
sleal-unity: This also should be internal, yep. Good catch.

com.unity.perception/Runtime/GroundTruth/PerceptionCamera.cs
sleal-unity: Good call-out, I jumped the gun by making a few too many changes to this file while poking around, and so I thought that I might as well go all in and get it done now instead of half here and then another half in another PR.

com.unity.perception/Runtime/GroundTruth/PerceptionUpdater.cs
```diff
@@ -0,0 +1,24 @@
...
+        {
+            var updaterObject = new GameObject("PerceptionUpdater");
+            updaterObject.AddComponent<PerceptionUpdater>();
+            updaterObject.hideFlags = HideFlags.HideAndDontSave;
+            DontDestroyOnLoad(updaterObject);
+        }
+
+        void LateUpdate()
+        {
+            LabelManager.singleton.RegisterPendingLabels();
```
JonathanHUnity, line 20: This would cause undefined behavior with object creation/deletion in other LateUpdate() scripts. To avoid this causing user confusion, we should instead use [SetPlayerLoop](https://docs.unity3d.com/ScriptReference/LowLevel.PlayerLoop.SetPlayerLoop.html) to avoid this entirely.

sleal-unity: Good point. Another option I've used for the ScenarioBase class is the DefaultExecutionOrder attribute. If I add the following attribute configuration to the LabelManager class, it should always run after all other default MonoBehaviour implemenations:
[DefaultExecutionOrder(1)]

com.unity.perception/CHANGELOG.md
```diff
@@ -51,6 +51,8 @@ The randomizer methods OnCreate(), OnStartRunning(), and OnStopRunning() are now
 
 ### Removed
 
+Removed the entities package dependency
```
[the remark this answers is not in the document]
sleal-unity: Can do.

com.unity.perception/Runtime/GroundTruth/Labeling/LabelManager.cs
JonathanHUnity: Linked lists provide no cache coherence and extra GC allocations, so it shouldn't take a deep dive to see that this is an inefficient data structure. I am not going to fight on this, since I realize we have some much bigger performance issues, but I don't think it is wise to treat poorly performing code as a "will not cause issues" kind of problem, since poor performance is generally a death by a thousand cuts kind of problem.

One thing that would make this structure more efficient would be to swap out the LinkedList<T> for a LinkedList<List<T>>, where you allocate in chunks to amortize the GC allocations and memory jumping while still maintaining O(1) on other operations.

sleal-unity: My point with the AddComponentObject API is to prove that we didn't necessarily have cache coherence in the first place (new chunk allocation required per new entity for SharedComponents), so the new LinkedHashSet solution is likely equivalent or faster than our previous implementation due to reduced ECS overhead. In this sense, the new data structure should have the same optimization prioritization as the former ECS solution (which wasn't even on our radar IIRC).

In regards to your suggestion, do you have a link to an existing implementation of this data structure I could reference? It sounds very neat.

com.unity.perception/Runtime/GroundTruth/Labeling/LabelManager.cs
```diff
@@ -0,0 +1,168 @@
...
+        /// <summary>
+        /// Refresh ground truth generation for the labeling of a particular GameObject. This is necessary when the
+        /// list of labels changes or when renderers or materials change on objects in the hierarchy.
+        /// </summary>
+        /// <param name="labeledObject">the component to refresh</param>
+        internal void RefreshLabeling(Labeling labeledObject)
+        {
+            m_RegisteredLabels.Remove(labeledObject);
+            m_LabelsPendingRegistration.Add(labeledObject);
+        }
```
mkamalza, line 112: perhaps rename `labeledObject` to `labeling` to better reflect the type of the function arguments?

com.unity.perception/Runtime/GroundTruth/PerceptionUpdater.cs
mkamalza: This caught my attention too. Let's make sure we don't get undefined behaviour here. Apart from the attribute, we could also use [this](https://docs.unity3d.com/Manual/class-MonoManager.html) window to set the execution order to sth very high.

com.unity.perception/Runtime/GroundTruth/SimulationState.cs
```diff
@@ -674,7 +667,7 @@ public AsyncAnnotation ReportAnnotationAsync(AnnotationDefinition annotationDefi
             foreach (var value in values)
                 jArray.Add(new JRaw(DatasetJsonUtility.ToJToken(value)));
 
-            ReportAsyncAnnotationResult<T>(asyncAnnotation, filename, jArray);
+            ReportAsyncAnnotationResult(asyncAnnotation, filename, jArray);
         }
 
         [SuppressMessage("ReSharper", "PossibleInvalidOperationException")]
```
mkamalza, line 673: It looks like QuaternionConvertor and Vector3Convertor are no longer needed. The only other usage of them is in tests. These were used for creating a specific output format but it later the code was changed to just use unity's native serialization.

mkamalza (changes requested):
Great work!

com.unity.perception/Runtime/GroundTruth/Labeling/LabelManager.cs
```diff
@@ -0,0 +1,167 @@
...
+        /// Recursively initializes Renderer components on GameObjects with Labeling components
+        /// </summary>
+        /// <param name="gameObject">The parent GameObject being initialized</param>
+        /// <param name="mpb">A reusable material property block</param>
+        /// <param name="labeling">The labeling component attached to the current gameObject</param>
+        /// <param name="instanceId">The perception specific instanceId assigned to the current gameObject</param>
+        void RecursivelyInitializeGameObjects(
+            GameObject gameObject, MaterialPropertyBlock mpb, Labeling labeling, uint instanceId)
+        {
+            var terrain = gameObject.GetComponent<Terrain>();
```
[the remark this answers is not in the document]
mkamalza: p.s. I don't think this is necessary for the current PR. The code already behaved like this. Just sth to keep in mind.

com.unity.perception/Runtime/GroundTruth/Labeling/LabelManager.cs
JonathanHUnity: I hear you - that's a good point. Here is an idea for your consideration: https://www.jacksondunstan.com/articles/4963
There are lots of ways to improve this system honestly, but I agree we need to take some extra time to come back to performance as a whole on this type of project.

com.unity.perception/Runtime/GroundTruth/Labelers/SemanticSegmentationLabeler.cs
JonathanHUnity: hmm, alrighty then :)

com.unity.perception/Runtime/GroundTruth/PerceptionUpdater.cs
sleal-unity: The attribute I suggested is equivalent to manipulating the script's position using this window. As for setting the value very high, I'll make sure to set it reasonably high so that the average user won't author any scripts that will interfere with this updater, but there are other default Unity processes that have low priority values, so I'll probably avoid going past those.

com.unity.perception/Runtime/GroundTruth/Labeling/LabelManager.cs
sleal-unity: done

com.unity.perception/Runtime/GroundTruth/Labeling/LabelManager.cs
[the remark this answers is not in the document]
sleal-unity: I will leave this optimization for another PR if that's alright.

mkamalza (approved):

JonathanHUnity (approved):

PR #241, opened 2021-03-09
Fix wrong labels on 2d bounding boxes

JonathanHUnity:
# Peer Review Information:
Label ids and names were incorrect on 2d bounding boxes when visualizations were disabled and all labeled objects are deleted each frame. This was being caused by a bug with our use of LabelEntryMatchCache, which updates itself whenever objects are assigned instance ids, causing the entries in the cache to report current instanceId->label mappings even when processing bounding boxes from multiple frames in the past.

This significant set of changes enables creating copies of the LabelEntryMatchCache each frame to freeze the state of the label mappings for use when the bounding boxes are calculated. It also:

* Significantly simplifies RenderTextureReader, forcing the user of this internal class to call Capture() whenever a capture is required
* Changed OnBeginRendering and OnEndRendering to include a ScriptableRenderContext parameter to allow for rendering code.
* Enables testing of 2d bounding box labeler directly

## Dev Testing:
**Tests Added**: Added test for 2d bounding box output in this specific case

**Core Scenario Tested**: Tested with custom project with ForegroundObjectPlacementRandomizer modified to no longer cache objects.

**At Risk Areas**: All ground truth, although output is well covered by automation.

**Notes + Expectations**: 

## Checklist
- [ ] - Updated docs
- [x] - Updated changelog


Wesley201 (approved):
Looks good and a huge refactor, tests additions are nice

com.unity.perception/Runtime/GroundTruth/Labeling/LabelEntryMatchCache.cs
```diff
@@ -55,10 +73,19 @@ void IGroundTruthGenerator.SetupMaterialProperties(MaterialPropertyBlock mpb, Re
             }
         }
 
+        /// <inheritdoc/>
```
sleal-unity, line 76: Just a note: the package validation suite is complaining about the lack of an XML doc on IGroundTruthGenerator.SetupMaterialProperties in this class, even though technically the implementation of this interface is private (which is a thing I didn't even know you could do)

sleal-unity (changes requested):

sleal-unity (approved):

PR #248, opened 2021-03-12
Added ability to disable labeling on an object by disabling its Labeling component

mkamalza:
# Peer Review Information:

Added ability to switch labeling on/off on an object at runtime through enabling or disabling its Labeling component.

**Editor Version Target (i.e. 19.3, 20.1)**: 2020.2

## Dev Testing:
**Tests Added**: 
Yes
**Core Scenario Tested**: 

**At Risk Areas**: 

**Notes + Expectations**: 

I will update **changelog** at merge time 

## Checklist
- [ ] - Updated docs
- [ ] - Updated changelog


com.unity.perception/Runtime/GroundTruth/Labeling/Labeling.cs
```diff
@@ -34,14 +35,19 @@ public class Labeling : MonoBehaviour
...
-        void OnDestroy()
+        void OnEnable()
         {
-            labelManager.Unregister(this);
+            RefreshLabeling();
+        }
+
+        void OnDisable()
+        {
+            RefreshLabeling();
```
sleal-unity, line 50: Unless I'm mistaken, shouldn't we be unregistering the label on disable here vs. refreshing?

sleal-unity (changes requested):

com.unity.perception/Runtime/GroundTruth/Labeling/Labeling.cs
mkamalza: In order for the label to be effectively removed from the ground truth generators, it needs to go through the "registration" process again, which makes it go through SetupMaterialProperties again. This is were the enabled state is checked and if disabled, the ground truth generators are reset for the object.

I think about it like this: Every labeling component that exists is registered, similar to the entity query we had. But those that are disabled are ignored in ground truth generation.

sleal-unity: gotcha

sleal-unity (approved):

com.unity.perception/Runtime/GroundTruth/Labelers/KeypointLabeler.cs
```diff
@@ -332,20 +332,21 @@ bool DoesTemplateContainJoint(JointLabel jointLabel)
 
         void ProcessLabel(Labeling labeledEntity)
         {
-            // Cache out the data of a labeled game object the first time we see it, this will
-            // save performance each frame. Also checks to see if a labeled game object can be annotated.
-            if (!m_KnownStatus.ContainsKey(labeledEntity.instanceId))
+            if (idLabelConfig.TryGetLabelEntryFromInstanceId(labeledEntity.instanceId, out var labelEntry))
```
JonathanHUnity, line 335: For future reference, avoid surrounding entire method bodies in 'if' statements. Instead, do `if (!xxx) return;`

com.unity.perception/Tests/Runtime/GroundTruthTests/LabelEntryMatchCacheTests.cs
```diff
@@ -131,5 +131,43 @@ public IEnumerator TryGet_ReturnsFalse_ForNonMatchingLabel_WhenAllObjectsAreDest
             }
         }
 
+        [UnityTest]
+        public IEnumerator TryGet_ReturnsFalse_ForMatchingLabelWithDisabledLabelingComponent()
```
JonathanHUnity, line 135: We should have a test for the keypoint labeler as well, since it does not use any of the tested codepaths.

JonathanHUnity (commented):
I realize this PR is already merged, but there are a few issues that I think should be resolved before release.

com.unity.perception/Runtime/GroundTruth/Labelers/KeypointLabeler.cs
mkamalza: I will change this. 
Is there a C# or Unity coding conventions reference specifically on this topic? (early return vs single return)

JonathanHUnity: nothing specific afaik - it's a more subjective assessment

PR #255, opened 2021-03-17
Amending dataset schema to describe correct coordinate system

mrpropellers:
The summary for the 3d bounding box states the coordinates are from the sensor's reference frame, but the string for the translational component says it's global.  I *think* this is just a typo in the doc, so correcting the description to match the summary for this type of entry.


StevenBorkman (approved):
Yeah, nice catch. That's definitely supposed to be the sensor's coordinate system

PR #274, opened 2021-03-26
Adding sky color control to semantic segmentation

JonathanHUnity:
# Peer Review Information:
This change adds a sky color to the semantic segmentation labeler.
Note that when this color is changed from the default black, it is no longer considered "background" by the visualizer. I think this is what users will expect, but open to feedback.

Before:
segmentation_3

After:
segmentation_4

## Dev Testing:
**Tests Added**: Added tests for sky coloring

**Core Scenario Tested**: Tested on URP and HDRP

**At Risk Areas**: 

## Checklist
- [ ] - Updated docs
- [ ] - Updated changelog
- [ ] - Updated test rail

### DIFF
--- a/TestProjects/PerceptionHDRP/Packages/packages-lock.json
+++ b/TestProjects/PerceptionHDRP/Packages/packages-lock.json
@@ -81,7 +81,7 @@
         "com.unity.collections": "0.9.0-preview.6",
         "com.unity.nuget.newtonsoft-json": "1.1.2",
         "com.unity.render-pipelines.core": "7.1.6",
-        "com.unity.simulation.capture": "0.0.10-preview.19",
+        "com.unity.simulation.capture": "0.0.10-preview.20",
         "com.unity.simulation.client": "0.0.10-preview.10",
         "com.unity.simulation.core": "0.0.10-preview.22"
       }
@@ -126,7 +126,7 @@
       "url": "https://packages.unity.com"
     },
     "com.unity.simulation.capture": {
-      "version": "0.0.10-preview.19",
+      "version": "0.0.10-preview.20",
       "depth": 1,
       "source": "registry",
       "dependencies": {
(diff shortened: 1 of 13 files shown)


com.unity.perception/Tests/Runtime/GroundTruthTests/SegmentationGroundTruthTests.cs
```diff
@@ -63,6 +63,7 @@ public class SegmentationPassTests : GroundTruthTestBase
     {
         static readonly Color32 k_SemanticPixelValue = new Color32(10, 20, 30, Byte.MaxValue);
         private static readonly Color32 k_InstanceSegmentationPixelValue = new Color32(255,0,0, 255);
+        private static readonly Color32 k_SkyValue = new Color32(10, 20, 30, 40);
```
StevenBorkman, line 66: Does the (10, 20, 30, 40) mean something?

TestProjects/PerceptionHDRP/Packages/packages-lock.json
```diff
@@ -81,7 +81,7 @@
         "com.unity.collections": "0.9.0-preview.6",
         "com.unity.nuget.newtonsoft-json": "1.1.2",
         "com.unity.render-pipelines.core": "7.1.6",
-        "com.unity.simulation.capture": "0.0.10-preview.19",
+        "com.unity.simulation.capture": "0.0.10-preview.20",
```
StevenBorkman, line 84: I don't have a problem with this, but just wanted to make sure that we wanted to upgrade capture to 20 in this fix, and not just happening by accident.

StevenBorkman (approved):
Nice job.

TestProjects/PerceptionHDRP/Packages/packages-lock.json
JonathanHUnity: Yea, good eye. this packages-lock.json is an auto-generated file that should have already been updated. It doesn't actually represent a change in the dependencies.

com.unity.perception/Tests/Runtime/GroundTruthTests/SegmentationGroundTruthTests.cs
JonathanHUnity: no, just wanted to use unique values for each channel to be super sure that we get the same data in as out.

PR #281, opened 2021-03-31
Added first frame skipping back to local editor runs

sleal-unity:
# Peer Review Information:
This PR adds the first frame check back to the PerceptionScenario to make sure we're skipping the first frame when generating datasets locally.

### DIFF
--- a/com.unity.perception/Runtime/Randomization/Scenarios/PerceptionScenario.cs
+++ b/com.unity.perception/Runtime/Randomization/Scenarios/PerceptionScenario.cs
@@ -20,7 +20,6 @@ namespace UnityEngine.Perception.Randomization.Scenarios
         /// </summary>
         MetricDefinition m_IterationMetricDefinition;
 
-
         /// <summary>
         /// Skip the first simulation frame if not running on Unity simulation
         /// </summary>
--- a/com.unity.perception/Runtime/Randomization/Scenarios/PerceptionScenario.cs
+++ b/com.unity.perception/Runtime/Randomization/Scenarios/PerceptionScenario.cs
@@ -20,6 +20,13 @@ namespace UnityEngine.Perception.Randomization.Scenarios
         /// </summary>
         MetricDefinition m_IterationMetricDefinition;
 
+
+        /// <summary>
+        /// Skip the first frame if not running on Unity simulation
+        /// </summary>
+        protected override bool isScenarioReadyToStart =>
+            !Configuration.Instance.IsSimulationRunningInCloud() && Time.frameCount == 1;
+
         /// <inheritdoc/>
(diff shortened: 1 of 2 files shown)


com.unity.perception/Runtime/Randomization/Scenarios/PerceptionScenario.cs
```diff
@@ -20,6 +20,12 @@ public abstract class PerceptionScenario<T> : Scenario<T> where T : ScenarioCons
         /// </summary>
         MetricDefinition m_IterationMetricDefinition;
 
+        /// <summary>
+        /// Skip the first simulation frame if not running on Unity simulation
+        /// </summary>
+        protected override bool isScenarioReadyToStart =>
```
JonathanHUnity, line 26: This should have a test. Also, do we need isScenarioReadyToStart if we can guarantee that all setup is synchronous? It feels like this logic should live in one of the base classes, since it is important for all simulations.

JonathanHUnity: Also, preferably this PR would target master, since it addresses a more generic problem. Then we can merge master back into this branch.

JonathanHUnity (changes requested):

com.unity.perception/Runtime/Randomization/Scenarios/PerceptionScenario.cs
sleal-unity: This code is already in master (though in a slightly different form) within the PerceptionScenario class. I think I removed this code from this branch by mistake at some point, so I'm adding it back in.

The reason why this is in the PerceptionScenario and not the ScenarioBase or other classes is because only perception based scenarios will run into this issue (and only when running locally at that). I've already seen Alex Zook remove this code from his custom derived Scenario class because he was running in sprawl and it was skipping the first frame when it shouldn't have. I don't want to put it too deep into the update loop code where a user couldn't remove it if need be. Which reminds me that I may need to revisit how to make the entire Scenario update loop more "overridable" considering the challenges we faced during hack week to get it work with ML agents.

Also, I originally designed isScenarioReadyToStart use cases where some global Scenario setup step has to occur over multiple frames. I can't give a concrete example of why a user would need to do this at the moment though, so I could probably pull it out if needed. I wouldn't know where to put the frame skipping function though if we still want it to be overrideable (the update loop can't be overrided at the moment).

sleal-unity: I just realized that we could make this change on the SynthDet side and just submit a more generic PR to master like you suggested. I'll go ahead and close this PR and make a new one targeting master later.

PR #285, opened 2021-04-01
Poisson disk super sampling

sleal-unity:
# Peer Review Information:
This PR addresses a problem Bowen found points sampled from the PoissonDiskSampling utility. Before this fix, the utility was generating a distribution of points that was biased toward the edges of the sampling region due to the over-fitting rejection-sampling edge-case behavior of the utility.

To fix this issue, the utility now samples a suitably larger point region and then crops it down to the original intended size to remove all points affected by this edge case behavior.

Finally, there was one final issue, and that was the initial point seeding. The first random point was being placed toward the center of the algorithm to compensate for making the occupancy grid too small (I generated the grid's dimensions with FloorToInt instead of CeilToInt lol). Now, the seed point is correctly placed in a completely random position within the generation region.

Before fix:
image
image

After the super sampling fix (I kinda messed up this histogram, but you get the picture):
image
image

And after the proper initial point seeding fix:
image
image

## Checklist
- [X] - Updated changelog

### DIFF
--- a/com.unity.perception/Runtime/Randomization/Randomizers/RandomizerExamples/Utilities/PoissonDiskSampling.cs
+++ b/com.unity.perception/Runtime/Randomization/Randomizers/RandomizerExamples/Utilities/PoissonDiskSampling.cs
@@ -147,9 +147,9 @@ namespace UnityEngine.Perception.Randomization.Randomizers.Utilities
 
             // Calculate occupancy grid dimensions
             var random = new Unity.Mathematics.Random(seed);
-            var cellSize = minimumRadius / math.sqrt(2);
-            var rows = Mathf.FloorToInt(height / cellSize);
-            var cols = Mathf.FloorToInt(width / cellSize);
+            var cellSize = minimumRadius / math.sqrt(2f);
+            var rows = Mathf.CeilToInt(height / cellSize);
+            var cols = Mathf.CeilToInt(width / cellSize);
             var gridSize = rows * cols;
             if (gridSize == 0)
                 return samples;
@@ -167,10 +167,8 @@ namespace UnityEngine.Perception.Randomization.Randomizers.Utilities
             // This list will track all points that may still have space around them for generating new points
             var activePoints = new NativeList<float2>(Allocator.Temp);
 
-            // Randomly place a seed point in a central location within the generation space to kick off the algorithm
-            var firstPoint = new float2(
-                random.NextFloat(0.4f, 0.6f) * width,
-                random.NextFloat(0.4f, 0.6f) * height);
+            // Randomly place a seed point to kick off the algorithm
+            var firstPoint = new float2(random.NextFloat(0f, width), random.NextFloat(0f, height));
             samples.Add(firstPoint);
             var firstPointCol = Mathf.FloorToInt(firstPoint.x / cellSize);
             var firstPointRow = Mathf.FloorToInt(firstPoint.y / cellSize);
--- a/com.unity.perception/Runtime/Randomization/Randomizers/RandomizerExamples/Utilities/PoissonDiskSampling.cs
+++ b/com.unity.perception/Runtime/Randomization/Randomizers/RandomizerExamples/Utilities/PoissonDiskSampling.cs
@@ -21,13 +21,15 @@ namespace UnityEngine.Perception.Randomization.Randomizers.Utilities
         /// <param name="minimumRadius">The minimum distance required between each sampled point</param>
         /// <param name="seed">The random seed used to initialize the algorithm state</param>
         /// <param name="samplingResolution">The number of potential points sampled around every valid point</param>
+        /// <param name="allocator">The allocator to use for the samples container</param>
         /// <returns>The list of generated poisson points</returns>
         public static NativeList<float2> GenerateSamples(
             float width,
             float height,
             float minimumRadius,
             uint seed = 12345,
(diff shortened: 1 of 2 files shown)


com.unity.perception/Runtime/Randomization/Randomizers/RandomizerExamples/Utilities/PoissonDiskSampling.cs
```diff
@@ -71,6 +85,44 @@ public void Execute()
...
+                        && point.y >= minimumRadius && point.y <= height + minimumRadius;
+                }
+
+                for (var i = 0; i < superSampledPoints.Length; i++)
+                {
+                    if (results[i])
+                    {
+                        var p = superSampledPoints[i];
+                        croppedSamples.Add(new float2(p.x - minimumRadius, p.y - minimumRadius));
+                    }
```
StevenBorkman, line 119: I don't understand why you are looping through this data twice. Couldn't this just be:

` for (var i = 0; i < superSampledPoints.Length; i++)
 {
     var point = superSampledPoints[i];
     if (point.x >= minimumRadius && point.x <= width + minimumRadius
           && point.y >= minimumRadius && point.y <= height + minimumRadius)
     {
         croppedSamples.Add(new float2(p.x - minimumRadius, p.y - minimumRadius));
     }          
}`


If you did that you don't have to allocated the results array.

StevenBorkman (changes requested):

com.unity.perception/Runtime/Randomization/Randomizers/RandomizerExamples/Utilities/PoissonDiskSampling.cs
sleal-unity: Ahh, so I separated this code into two pieces so that burst could vectorize the first loop and do less work on the un-vectorizable second loop. I'll annotate the code with this information.

sleal-unity: Technically, there's a multi-threaded way of splitting off this filtering work with the [IJobParallelForFilter](https://docs.unity3d.com/Packages/com.unity.jobs@0.0/api/Unity.Jobs.IJobParallelForFilter.html) interface, but I thought that was overkill.

sleal-unity: Here's a neat look at the burst inspector to see all the vectorization it does at the first loop (basically every purple instruction):
image

sleal-unity: In contrast, burst can't seem to vectorize any part of the second loop:
image

com.unity.perception/Runtime/Randomization/Randomizers/RandomizerExamples/Utilities/PoissonDiskSampling.cs
```diff
@@ -71,6 +85,51 @@ public void Execute()
...
+            [ReadOnly] public NativeArray<float2> superSampledPoints;
+            public NativeList<float2> croppedSamples;
+
+            public void Execute()
+            {
+                var results = new NativeArray<bool>(
+                    superSampledPoints.Length, Allocator.Temp, NativeArrayOptions.UninitializedMemory);
+
+                // The comparisons operations made in this loop are done separately from the list-building loop
+                // so that burst can automatically generate vectorized assembly code this portion of the job.
```
mkamalza, line 107:
```suggestion
                // so that burst can automatically generate vectorized assembly code for this portion of the job.
```

com.unity.perception/Runtime/Randomization/Randomizers/RandomizerExamples/Utilities/PoissonDiskSampling.cs
```diff
@@ -71,6 +85,51 @@ public void Execute()
...
+                for (var i = 0; i < superSampledPoints.Length; i++)
+                {
+                    if (results[i])
+                        croppedSamples.Add(superSampledPoints[i]);
+                }
+
+                // Remove the positional offset from the filtered-but-still-super-sampled points
+                for (var i = 0; i < croppedSamples.Length; i++)
+                {
+                    croppedSamples[i] -= new float2(minimumRadius, minimumRadius);
```
mkamalza, line 126: Can we create this float2 just once instead of in each iteration of the loop, or would that mess up the burst compiler?

mkamalza (approved):

com.unity.perception/Runtime/Randomization/Randomizers/RandomizerExamples/Utilities/PoissonDiskSampling.cs
sleal-unity: Yep. I moved it outside the loop for ya.

StevenBorkman (approved):

com.unity.perception/Runtime/Randomization/Randomizers/RandomizerExamples/Utilities/PoissonDiskSampling.cs
StevenBorkman: Interesting. I figured you had a good reason for it

PR #290, opened 2021-04-13
Fixing keypoint bugs when visualizations are disabled

JonathanHUnity:
# Peer Review Information:
Keypoints were being mislabeled when visualizations were disabled due to the state of the keypoints being overwritten by frames happening in between simulation and readback.

This change makes a copy of the data before storing it for future use on readback.

## Dev Testing:
**Tests Added**:  Added test for new case

**Core Scenario Tested**: Tested keypoints locally and on the PeopleSansPeople project

**At Risk Areas**: 

## Checklist
- [ ] - Updated docs
- [ ] - Updated changelog
- [ ] - Updated test rail

### DIFF
--- a/com.unity.perception/Runtime/GroundTruth/Labelers/KeypointLabeler.cs
+++ b/com.unity.perception/Runtime/GroundTruth/Labelers/KeypointLabeler.cs
@@ -177,8 +177,9 @@ namespace UnityEngine.Perception.GroundTruth
             {
                 if (InstanceIdToColorMapping.TryGetColorFromInstanceId(keypointSet.Key, out var idColor))
                 {
-                    foreach (var keypoint in keypointSet.Value.keypoints)
+                    for (var i = 0; i < keypointSet.Value.keypoints.Length; i++)
                     {
+                        var keypoint = keypointSet.Value.keypoints[i];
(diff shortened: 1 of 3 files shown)


com.unity.perception/Runtime/GroundTruth/Labelers/KeypointLabeler.cs
```diff
@@ -458,7 +461,18 @@ void ProcessLabel(Labeling labeledEntity)
...
+                var keypointEntry = new KeypointEntry()
+                {
+                    instance_id = cachedKeypointEntry.instance_id,
+                    keypoints = cachedKeypointEntry.keypoints.ToArray(),
+                    label_id = cachedKeypointEntry.label_id,
+                    pose = cachedKeypointEntry.pose,
+                    template_guid = cachedKeypointEntry.template_guid
+                };
+                m_AsyncAnnotations[m_CurrentFrame].keypoints[labeledEntity.instanceId] =
+                    /*cachedKeypointEntry;*/ keypointEntry;
```
StevenBorkman, line 475: You left some commented out code in here

StevenBorkman (changes requested):
Good catch. Yet another thing that we have to be intentional when testing, making sure that everything works with visuals off.

StevenBorkman (approved):

PR #294, opened 2021-04-14
Supporting the 'step' button in the editor

JonathanHUnity:
# Peer Review Information:
This change allows PerceptionCameras to function properly when using the 'step' button in the editor.

The labelers in the PerceptionCamera previously were not even running when using the 'step' button because the editor forces game time to increment at exactly 20ms per frame when using this button. This messed up the timing code in DatasetCapture.

This change detects the use of the step button and forces all sensors to start render and capture each frame. This will not work great with the 'frames between capture' feature, so open to feedback on that one.

## Dev Testing:
**Tests Added**: Added editor test for dataset capture while stpping

**Core Scenario Tested**: Tested in editor

**At Risk Areas**: 

## Checklist
- [ ] - Updated docs
- [x] - Updated changelog
- [ ] - Updated test rail

### DIFF
--- a/com.unity.perception/Runtime/GroundTruth/SimulationState.cs
+++ b/com.unity.perception/Runtime/GroundTruth/SimulationState.cs
@@ -452,6 +452,8 @@ namespace UnityEngine.Perception.GroundTruth
                 {
                     //When the user clicks the 'step' button in the editor, frames will always progress at .02 seconds per step.
                     //In this case, just run all sensors each frame to allow for debugging
+                    Debug.Log($"Frame step forced all sensors to synchronize, changing frame timings.");
+
                     sensorData.sequenceTimeOfNextRender = UnscaledSequenceTime;
                     sensorData.sequenceTimeOfNextCapture = UnscaledSequenceTime;
                 }
--- a/com.unity.perception/Runtime/GroundTruth/SimulationState.cs
+++ b/com.unity.perception/Runtime/GroundTruth/SimulationState.cs
@@ -447,6 +447,16 @@ namespace UnityEngine.Perception.GroundTruth
             {
                 var sensorData = m_Sensors[activeSensor];
 
+#if UNITY_EDITOR
+                if (UnityEditor.EditorApplication.isPaused)
+                {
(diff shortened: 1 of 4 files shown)


com.unity.perception/Runtime/GroundTruth/SimulationState.cs
```diff
@@ -447,6 +447,16 @@ public void Update()
             {
                 var sensorData = m_Sensors[activeSensor];
 
+#if UNITY_EDITOR
+                if (UnityEditor.EditorApplication.isPaused)
+                {
+                    //When the user clicks the 'step' button in the editor, frames will always progress at .02 seconds per step.
+                    //In this case, just run all sensors each frame to allow for debugging
+                    sensorData.sequenceTimeOfNextRender = UnscaledSequenceTime;
+                    sensorData.sequenceTimeOfNextCapture = UnscaledSequenceTime;
```
mkamalza, line 456: Besides the issues you mentioned in the PR description, this also causes a timing issue in multi-sensor simulations where the sensors have different delta times. This code will cause all sensors to capture at the same time regardless of their delta times. It would be nice if we could distinguish between pause and step, and only do this if step is used.

I can't think of an immediate solution though, and being able to use step is really important. So it may be worthwhile to just add some kind of warning message to the screen (or the log) when the player is paused, to inform the user that timing settings will not be respected if the simulation is paused?

mkamalza (changes requested):
Very nice to see step working!

com.unity.perception/Runtime/GroundTruth/SimulationState.cs
JonathanHUnity: As far as I can tell, this code will only get hit if the user hits the step button. When the user pauses and unpauses it will never get hit. I could add an info message to the log mentioning this behavior - does that sound good?

mkamalza: My bad, you're right it only gets called on step!
Yeah log message sounds good.

JonathanHUnity: Done!

mkamalza (approved):

PR #304, opened 2021-04-27
added uint field element for random seeds

sleal-unity:
# Peer Review Information:
This PR adds a new UIntField UI element to enable users to configure the full range of random-seeds possible for their scenarios. Before this PR, the scenario's random seed constant (stored as an unsigned integer) was represented by an integer field that limited the potential seed values that could be assigned to the scenario and also introduced serialization issues when negative seed values were entered.

In addition, a randomize-seed button was also added to the Run in Unity Simulation window so users can quickly modify their scenario's random seed between runs.

## Dev Testing:
No tests were added, though I will need to update the test rail to confirm the changes made in this PR.

## Checklist
- [X] - Updated changelog
- [ ] - Updated test rail

### DIFF
--- a/com.unity.perception/CHANGELOG.md
+++ b/com.unity.perception/CHANGELOG.md
@@ -14,6 +14,8 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 ### Added
 Added support for 'step' button in editor.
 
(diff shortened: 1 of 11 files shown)


mkamalza (approved):
Great stuff. I added an additional comment on line 19, please check if it makes sense.

PR #308, opened 2021-05-05
Generic Asset Sources

sleal-unity:
# Peer Review Information:
This PR introduces the following new content to the perception package:
- Generic AssetSources class for abstracting the loading of assets from within Randomizers
- Archetypes for preprocessing assets loaded from AssetSources
- A bunch of UI elements to the created the associated inspector UIs for AssetSources and archetypes

## Dev Testing:
**Tests Added**: 

**Core Scenario Tested**: 

**At Risk Areas**: 

## Checklist
- [ ] - Updated docs
- [X] - Updated changelog
- [ ] - Updated test rail


com.unity.perception/Runtime/Randomization/Randomizers/RandomizerExamples/Utilities/GameObjectOneWayCache.cs
```diff
@@ -13,6 +14,8 @@ public class GameObjectOneWayCache
     {
         static ProfilerMarker s_ResetAllObjectsMarker = new ProfilerMarker("ResetAllObjects");
 
+        GameObject[] m_Prefabs;
```
JonathanHUnity, line 17: Why is this class changing?

com.unity.perception/Runtime/Randomization/Randomizers/AssetSources/AssetSource.cs
```diff
@@ -0,0 +1,139 @@
...
+using UnityEngine.Perception.Randomization.Samplers;
+
+namespace UnityEngine.Perception.Randomization
+{
+    /// <summary>
+    /// AssetSources are used to load assets from a generically within a <see cref="Randomizers.Randomizer"/>
+    /// </summary>
+    /// <typeparam name="T">The type of asset to load</typeparam>
+    [Serializable]
+    public sealed class AssetSource<T> where T : Object
```
JonathanHUnity, line 11: Pretty significant new feature here - there should also be tests and docs. Alternately, we could keep it on a branch and move it into a 'beta' period in which we put it through its paces.

com.unity.perception/Runtime/Randomization/Randomizers/AssetSources/AssetSource.cs
```diff
@@ -0,0 +1,139 @@
...
+        {
+            CheckIfInitialized();
+            return assetSourceLocation.LoadAsset<T>(index);
+        }
+
+        /// <summary>
+        /// Returns all unprocessed assets that can be loaded from this AssetSource
+        /// </summary>
+        /// <returns>All assets that can be loaded from this AssetSource</returns>
+        public T[] LoadAllRawAssets()
```
JonathanHUnity, line 61: It says "Returns all unprocessed assets" but it calls `archetype.Preprocess`. Intended?

com.unity.perception/Runtime/Randomization/Randomizers/AssetSources/AssetSource.cs
```diff
@@ -0,0 +1,139 @@
...
+            for (var i = 0; i < Count; i++)
+                array[i] = CreateProcessedInstance(i);
+            return array;
+        }
+
+        /// <summary>
+        /// Returns a uniformly random sampled asset from this AssetSource
+        /// </summary>
+        /// <returns>The randomly sampled asset</returns>
+        public T SampleAsset()
```
JonathanHUnity, line 100: Why randomize in this class? Randomization logic appears to be outside the scope of this class.

com.unity.perception/Runtime/Randomization/Randomizers/AssetSources/AssetSource.cs
```diff
@@ -0,0 +1,139 @@
...
+        public Archetype<T> archetype
+        {
+            get => (Archetype<T>)m_ArchetypeBase;
+            set => m_ArchetypeBase = value;
+        }
+
+        /// <summary>
+        /// The number of assets available within this asset source
+        /// </summary>
+        public int Count => assetSourceLocation.Count;
```
JonathanHUnity, line 35: properties should be camelCase (first letter lower)

com.unity.perception/Runtime/Randomization/Randomizers/AssetSources/AssetSource.cs
```diff
@@ -0,0 +1,139 @@
...
+        /// </summary>
+        public void ReleaseAssets()
+        {
+            CheckIfInitialized();
+            assetSourceLocation.ReleaseAssets();
+        }
+
+        void CheckIfInitialized()
+        {
+            if (!m_Initialized)
```
JonathanHUnity, line 127: Could this CheckIfInitialized call be in the AssetSourceLocation itself?

JonathanHUnity (changes requested):
Here are a few comments. More to come once I figure out how to use it.

com.unity.perception/Runtime/Randomization/Randomizers/RandomizerExamples/Utilities/GameObjectOneWayCache.cs
sleal-unity: This is a consequence of not using archetypes to preprocess prefabs directly. In the editor, modifications made to a prefab at runtime are still permanent, so I instantiate prefabs loaded by asset sources first before I modify them. Since now you have these new archetype processed instances that need to be tracked in your project, I thought managing them using the GameObjectCache would be easier than tracking a whole separate list of spawned instances at runtime.

Partly as a consequence of this new relationship, I needed a way to use GameObject caching that duplicates sourced instances instead of sourced prefabs. The changes to the cache reflect these new requirements.

com.unity.perception/Runtime/Randomization/Randomizers/AssetSources/AssetSource.cs
sleal-unity: Tests and docs are certainly coming after the major PR changes have been approved ;)

Separate branch works too. I think that makes more sense considering that we'll probs still have to make a few changes based on our iteration with the services team as we integrate this into datamaker.

com.unity.perception/Runtime/Randomization/Randomizers/AssetSources/AssetSource.cs
sleal-unity: Oops, def not intended. Good catch.

com.unity.perception/Runtime/Randomization/Randomizers/AssetSources/AssetSource.cs
sleal-unity: I'm down to remove the sampling features, but I just felt that simple uniform sampling would be a common use case, similar to how users sample from GameObjectParameters.

com.unity.perception/Runtime/Randomization/Randomizers/AssetSources/AssetSource.cs
sleal-unity: I agree, but I didn't want to break convention with the existing list class in C#. I've got no problem with switching it though.

com.unity.perception/Runtime/Randomization/Randomizers/AssetSources/AssetSource.cs
sleal-unity: I did this originally, but I moved this to the AssetSource class so that people implementing asset sources wouldn't have to duplicate the manual process of checking if the asset source has been initialized before sampling assets.

I can move it back though.

com.unity.perception/Runtime/Randomization/Randomizers/AssetSources/AssetSource.cs
sleal-unity: I've removed the randomization features

com.unity.perception/Runtime/Randomization/Randomizers/AssetSources/AssetSource.cs
sleal-unity: Counts have been changed to count

com.unity.perception/Runtime/Randomization/Randomizers/AssetSources/AssetSource.cs
sleal-unity: I've decided to leave this in the AssetSource

JonathanHUnity (changes requested):
We should keep this open as a branch in the beta period while testing in a few projects. Is there a new-design doc for this yet? We should migrate our design to this so that we can be explicit about the exit criteria from beta.

Also, my concern with this design that we need to validate in beta is the amount of work required to upgrade the project and how one is supposed to set up a project that works both in datamaker and locally. It might seem pedantic when it doesn't feel like much work, but this is a critical flow to get right as we start thinking about "up-selling to datamaker" https://unity.slack.com/archives/GTC85DTT8/p1623262598135700?thread_ts=1623261703.134700&cid=GTC85DTT8

com.unity.perception/Runtime/Randomization/Randomizers/AssetSources/AssetSourceLocationNotes.cs
```diff
@@ -0,0 +1,24 @@
+﻿using System;
+
+namespace UnityEngine.Perception.Randomization
+{
+    /// <summary>
+    /// Used to annotate an AssetSourceLocation with notes within the inspector UI
+    /// </summary>
+    public class AssetSourceLocationNotes : Attribute
```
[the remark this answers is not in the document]
sleal-unity: I've renamed this class to "AssetSourceEditorDescriptionAttribute"

com.unity.perception/Runtime/GroundTruth/Labelers/CameraLabeler.cs
```diff
@@ -103,6 +103,27 @@ public abstract class CameraLabeler
         /// </summary>
         protected virtual void Cleanup() {}
 
+        /// <summary>
+        /// Initializes labeler with the target perception camera
+        /// </summary>
+        /// <param name="camera">The target perception camera</param>
+        public void Init(PerceptionCamera camera)
```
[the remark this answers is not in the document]
mkamalza: This needed to be public in an earlier version of the dynamic label config creation flow, where the label config was built later than when the labelers were first initialized. It no longer needs to be public; will change back to internal.

com.unity.perception/Runtime/Randomization/Randomizers/AssetSources/AssetRoleBase.cs
```diff
@@ -0,0 +1,15 @@
+﻿using UnityEngine;
+
+namespace UnityEngine.Perception.Randomization
+{
+    /// <summary>
+    /// The base asset role class. Derive from <see cref="AssetRole{T}"/> instead to create a new asset role.
+    /// </summary>
+    public abstract class AssetRoleBase
```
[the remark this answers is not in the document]
sleal-unity: Good idea. I'll do that.

com.unity.perception/Tests/Runtime/Randomization/ScenarioTests/ScenarioTests.cs
```diff
@@ -75,11 +75,11 @@ public void ScenarioConfigurationOverwrittenDuringDeserialization()
 
             // Serialize some values
             m_Scenario.constants = constants;
-            var serializedConfig = m_Scenario.SerializeToJson();
+            m_Scenario.configuration = new TextAsset(m_Scenario.SerializeToJson());
 
             // Change the values
             m_Scenario.constants = changedConstants;
-            m_Scenario.DeserializeFromJson(serializedConfig);
+            m_Scenario.DeserializeConfiguration();
```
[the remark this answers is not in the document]
sleal-unity: We don't necessarily want to de-serialize the config at will, we just want to differentiate between config assignment and config de-serialization so an editor assigned config can be deserialized without "reassigning" it when the scenario awakes.

With this in mind, the changes I'll make include:
- Making the DeserializeConfiguration() method protected so it can still be overridden by the UnitySimulationScenario class while removing it from the public interface
- Adding a DeserializeConfigurationInternal() method so we can continue to de-serialize configs for editor and testing purposes

mkamalza (approved):

JonathanHUnity (approved):

PR #312, opened 2021-05-06
Added option for foreground object bounds normalization

mkamalza:
# Peer Review Information:

This PR adds an option to the sample foreground object placement Randomizer to normalize the bounds of foreground objects. This is the same approach used in original SynthDet and much of the code has been copied from there.

**Editor Version Target**: 2020.3

## Dev Testing:
**Tests Added**: 

**Core Scenario Tested**: 

**At Risk Areas**: 

## Checklist
- [ ] - Updated docs
- [ ] - Updated changelog
- [ ] - Updated test rail


mkamalza:
We now have some indication of how this could affect model performance. For SynthDet, Bowen's tests have shown a 10% decrease in performance with the option enabled. It's important to note that I did modify the scale randomization to get similarly sized objects than before the change, otherwise the objects would be huge in the camera. That said, the relative scale of objects does change significantly compared to what we had before.

mkamalza:
We can include this in the Randomizer library but I don't think the default sample randomizer needs it

PR #315, opened 2021-05-19
Output directory

StevenBorkman:
# Peer Review Information:

## Dev Testing:
**Tests Added**: 

**Core Scenario Tested**: 

**At Risk Areas**: 

## Checklist
- [na] - Updated docs
- [x] - Updated changelog
- [x] - Updated test rail

### DIFF
--- a/com.unity.perception/Editor/GroundTruth/PerceptionCameraEditor.cs
+++ b/com.unity.perception/Editor/GroundTruth/PerceptionCameraEditor.cs
@@ -1,5 +1,6 @@
 ﻿using System;
 using System.Collections.Generic;
+using System.IO;
 using UnityEditorInternal;
 using UnityEngine;
 using UnityEngine.Perception.GroundTruth;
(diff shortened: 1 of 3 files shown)


mkamalza (approved):
Sweet!

StevenBorkman:
Added the following test to the testrail
image

com.unity.perception/Editor/GroundTruth/PerceptionCameraEditor.cs
```diff
@@ -151,6 +159,37 @@ public override void OnInspectorGUI()
                 GUILayout.EndVertical();
             }
 
+            GUILayout.Space(10);
+
+            var userBaseDir = PlayerPrefs.GetString(SimulationState.userBaseDirectoryKey);
```
JonathanHUnity, line 164: This path only needs to be set once per project, so it feels like it is using up a lot of space in the inspector. Also it is a project-wide setting. For both those reasons, should this be moved into a project settings window? See https://docs.unity3d.com/2019.1/Documentation/ScriptReference/SettingsProvider.html

With sensors coming soon (tm), PerceptionCamera will no longer be the single interface for dataset generation

JonathanHUnity (approved):
Yay! One note about the interface

com.unity.perception/Editor/GroundTruth/PerceptionCameraEditor.cs
mkamalza: Interesting. So this gives us the ability to have entries in the actual project settings window of unity?
I agree it would be nice to have this there. I think we should start thinking more seriously about taking some settings out of perception camera (e.g. simulation delta time) and putting them in a more central place once sensors start trickling in.

mkamalza:
Merging this for now and adding a JIRA task to revisit where the option is once sensors are added.

PR #344, opened 2021-07-06
Add camera projection information to metadata

leopoldo-zugasti:
# Peer Review Information:
Added projection field in the capture.sensor metadata to keep track of whether the capture was done in perspective or orthographic mode

## Dev Testing:
**Tests Added**: 
 Tests changed: 
```
[TestCase(1, "1")]
[TestCase(1u, "1")]
[TestCase(1.0, "1")]
[TestCase(1.0f, "1")]
[TestCase("string", "string")] // (Previously: [TestCase("string", "\"string\"")]) 
public void Primitive_ReturnsValue(object o, string jsonExpected)
{
      var jsonActual = DatasetJsonUtility.ToJToken(o).ToString();
      Assert.AreEqual(TestHelper.NormalizeJson(jsonExpected), 
      TestHelper.NormalizeJson(jsonActual));
}
```
**Core Scenario Tested**: 

**At Risk Areas**: 

## Checklist
- [x] - Updated docs
- [x] - Updated changelog
- [ ] - Updated test rail


com.unity.perception/Documentation~/Schema/Synthetic_Dataset_Schema.md
```diff
@@ -140,6 +140,7 @@ sensor {
   translation:      <float, float, float> -- Position in meters: (x, y, z) with respect to the ego coordinate system. This is typically fixed during the simulation, but we can allow small variation for domain randomization.
   rotation:         <float, float, float, float> -- Orientation as quaternion: (w, x, y, z) with respect to ego coordinate system. This is typically fixed during the simulation, but we can allow small variation for domain randomization.
   camera_intrinsic: <3x3 float matrix> [optional] -- Intrinsic camera calibration. Empty for sensors that are not cameras.
+  projection:       <string> -- holds the type of projection the camera used for that capture: Options: "perspective" or "orthographic"
```
mkamalza, line 143: We don't record this per capture do we? So maybe we should just say:
The camera's projection type. Options: "perspective" or "orthographic"

mkamalza (approved):
sweet

com.unity.perception/Documentation~/Schema/Synthetic_Dataset_Schema.md
leopoldo-zugasti: Actually we do, I don't know if you remember but we discussed were to put this field and we decided that since it is possible that the user changes the projection type throughout different captures we would place it inside the sensor capture data

leopoldo-zugasti: It also works well like that it is in the same place as the camera_intrinsic matrix for that capture
image

leopoldo-zugasti: I don't mind changing to what you said though

mkamalza: you're right, old age is catching up with me :)

leopoldo-zugasti: All the other fields there are also per capture but they don't mention "for this capture" in their description so what you said makes more sense regardless

PR #359, opened 2021-08-09
tutorial updates

mkamalza:
- Clarified the note about MSAA and empty images
- Small restructuring in Phase 2 of the tutorial to put the step for adding the randomizer after creating the randomizer tag


aryan-mann (approved):
Looks good 👍 

The only thing is that the main camera might not be called Main Camera in the scene, but that is most likely a tiny tiny subset of probably advanced users.

mkamalza:
> Looks good 👍
> 
> The only thing is that the main camera might not be called Main Camera in the scene, but that is most likely a tiny tiny subset of probably advanced users.

Since the tutorial uses a new scene that the user creates, we should be good as long as Unity editor uses Main Camera as the name for the camera already present in a new scene by default.

PR #380, opened 2021-09-29
fix perception binaries capture one extra frame before the scenario s…

RuiyuZ:
…tarts

# Peer Review Information:

## Dev Testing:
**Tests Added**: 

**Core Scenario Tested**: 

**At Risk Areas**: 

## Checklist
- [ ] - Updated docs
- [ ] - Updated changelog
- [ ] - Updated test rail

### DIFF
--- a/com.unity.perception/Runtime/Randomization/Scenarios/PerceptionScenario.cs
+++ b/com.unity.perception/Runtime/Randomization/Scenarios/PerceptionScenario.cs
@@ -20,29 +20,13 @@ namespace UnityEngine.Perception.Randomization.Scenarios
         /// </summary>
         MetricDefinition m_IterationMetricDefinition;
 
-        /// <summary>
-        /// The scriptable render pipeline hook used to capture perception data skips the first frame of the simulation
-        /// when running locally, so this flag is used to track whether the first frame has been skipped yet.
-        /// </summary>
-        protected bool m_SkippedFirstFrame;
-
-        /// <inheritdoc/>
-        protected override bool isScenarioReadyToStart
-        {
-            get
-            {
-                if (!m_SkippedFirstFrame)
-                {
-                    m_SkippedFirstFrame = true;
-                    return false;
-                }
-                return true;
-            }
-        }
+        protected override bool isScenarioReadyToStart => Time.frameCount >= 2;
 
         /// <inheritdoc/>
         protected override void OnAwake()
         {
+            foreach (var perceptionCamera in FindObjectsOfType<PerceptionCamera>())
+                perceptionCamera.enabled = false;
             m_IterationMetricDefinition = DatasetCapture.RegisterMetricDefinition(
                 "scenario_iteration", "Iteration information for dataset sequences",
                 Guid.Parse(k_ScenarioIterationMetricDefinitionId));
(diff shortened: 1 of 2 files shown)


com.unity.perception/Runtime/Randomization/Scenarios/PerceptionScenario.cs
```diff
@@ -20,29 +20,13 @@ public abstract class PerceptionScenario<T> : Scenario<T> where T : ScenarioCons
...
-            {
-                if (!m_SkippedFirstFrame)
-                {
-                    m_SkippedFirstFrame = true;
-                    return false;
-                }
-                return true;
-            }
-        }
+        protected override bool isScenarioReadyToStart => Time.frameCount >= 2;
```
JonathanHUnity, line 23: What is the original purpose of waiting until frame 2? In the standalone player is the first frame black or is there content visible? If content is visible, then we don't actually need to skip it. If content is not visible, we need to make sure that frame counting in SimulationState also skips this frame.

com.unity.perception/Runtime/Randomization/Scenarios/PerceptionScenario.cs
```diff
@@ -20,29 +20,13 @@ public abstract class PerceptionScenario<T> : Scenario<T> where T : ScenarioCons
...
 
         /// <inheritdoc/>
         protected override void OnAwake()
         {
+            foreach (var perceptionCamera in FindObjectsOfType<PerceptionCamera>())
```
JonathanHUnity, line 28: This works for any apps using Scenario, but this should also be addressed when not using a Scenario.

JonathanHUnity (changes requested):
Can you think of a way of testing this?

com.unity.perception/Runtime/Randomization/Scenarios/PerceptionScenario.cs
sleal-unity: I think the underlying issue was that scenarios were not starting at the same time the perception camera starts capturing. If a user is not using a scenario, then the issue is no longer valid. I'm not sure I understand your feedback in this case.

com.unity.perception/Runtime/Randomization/Scenarios/PerceptionScenario.cs
sleal-unity: The purpose of waiting until frame 2 is to synchronize scenario execution and perception camera capture behavior across all execution platforms (editor, player, cloud, etc.). In the editor, the perception camera only starts capturing on frame 2 when set to scheduled capture mode. However, in local players and cloud execution, the camera starts at frame 1, causing an extra frame to be captured before the scenario starts.

Syncing frame counting in the SimulationState is a good callout.

RuiyuZ: The content is visible in a standalone player, and yes as Steven said the purpose is to make the perception camera capture behave consistently across all execution platforms. So shall we still make this skipping frame 1 or leave it as it was?

JonathanHUnity: @sleal-unity are you saying that in the standalone player, the camera runs before the first Update?

com.unity.perception/Runtime/Randomization/Scenarios/PerceptionScenario.cs
JonathanHUnity: This code is disabling the perception camera. If it's the scenario timing that needs adjusting, we shouldn't need to touch the camera

PR #450, opened 2022-03-28
Fix instructions for finding generated dataset

csun:


com.unity.perception/Documentation~/Tutorial/Phase1.md
```diff
@@ -375,15 +375,9 @@ While the simulation is running, your _**Game**_ view will quickly generate fram
 
 ### <a name="step-7">Step 7: Inspect Generated Synthetic Data</a> 
 
-By default, the generated dataset will be generated in the Perception format. You can find the location of your data, and set the location of future datasets inside the project settings window. To achieve this: 
+By default, the generated dataset will be generated in the Perception format. You can find the location of your data, and set the location of future datasets in the settings on your Perception Camera object. To achieve this:
```
mkamalza, line 378:
```suggestion
By default, the output dataset will be generated in the Perception format. You can find the location of your data, and set the location of future datasets in the settings on your Perception Camera object. To achieve this: 
```

mkamalza: I'm curious, did we get the wires crossed here? Didn't we have the output folder in the project settings at some point?

mkamalza (changes requested):

com.unity.perception/Documentation~/Tutorial/Phase1.md
JonathanHUnity: The latest release (which we released last week) moved the directory into the project settings window. This looks like an issue of the user being on an old version.

csun: Ah, ok - so this PR is actually wrong. I've actually never used 0.10. I guess the real PR would be to fix the broken image in the docs showing the updated preferences window, which I can try to capture later

PR #475, opened 2022-04-21
Fix documentation and images

RuiyuZ:
* added an index.md which is the mirror of README.md, to serve as an entry point in the https://docs.hq.unity3d.com/Packages/com.unity.perception@0.11/
will fix the issue in this ticket: https://jira.unity3d.com/browse/AISV-2680
* move all the images used into /images directory, since the docs.hq.unity3d.com requires all the images to be placed under /images in order to show them correctly
will fix the issue in this ticket: https://jira.unity3d.com/browse/AISV-2552


JonathanHUnity (approved):
Great!

